PR #4, opened 2022-06-14
FS-1019: Restrict test deploys to main branch

gidsg:
Essential we do this for this repo so we don't run in branch migrations


harryyo (approved):
Looks great.

PR #5, opened 2022-06-14
FS 1022 update status

ramsharma-prog:
Adds a "status" column to the database and sets it to "UNASSESSED" by default & adds a new route to update the status.


SentientHousePlant (approved):
LGTM <3

db/models/assessment.py
```diff
@@ -59,3 +64,18 @@ def register_application(application_id: str):
...
+            assessment_id.compliance_status = status
+            db.session.commit()
+            assessment = Assessment(
+                application_id=assessment_id.application_id,
+                id=assessment_id.id,
+                compliance_status=assessment_id.compliance_status,
+            )
+            return assessment
+        except AttributeError:
+            raise AssessmentError(message="An assessment id doesn't not exist")
```
gidsg, line 81: typo
```suggestion
            raise AssessmentError(message="An assessment does not exist")
```

ramsharma-prog: Thank you 😄

db/models/assessment.py
```diff
@@ -59,3 +64,18 @@ def register_application(application_id: str):
...
             )
         return assessment
+
+    @staticmethod
+    def update_status(assessment_id: str, status: str):
+        try:
+            assessment_id = AssessmentMethods.get_by_id(assessment_id)
+            assessment_id.compliance_status = status
+            db.session.commit()
+            assessment = Assessment(
```
gidsg, line 74: Could we just return this directly rather than assigning it first?

db/models/assessment.py
gidsg: I think this try catch is not adding any value, as if the ID was not found it already would have thrown an exception here: https://github.com/communitiesuk/funding-service-design-assessment-store/blob/b84c92dc27864fd1749314fde47795fe4b84f738/db/models/assessment.py#L52

manifest.yml
```diff
@@ -13,4 +13,4 @@ applications:
   - https://github.com/cloudfoundry/python-buildpack.git
   command: flask run --host 0.0.0.0 --port 8080
   services:
-    - assessment-store-test-db 
\ No newline at end of file
+    - assessment-store-test-db
```
gidsg, line 16: This is just a whitespace change so this file could be reverted.

openapi/api.yml
```diff
@@ -87,3 +87,38 @@ paths:
...
+                schema:
+                  $ref: 'components.yml#/components/schemas/Assessment'
+        404:
+          description: ERROR - Assessment not found
+          content:
+            application/json:
+              schema:
+                $ref: 'components.yml#/components/schemas/GeneralError'
+              example:
+                code: 404
```
gidsg, line 122: Have we tested this actually returns a 404? I think it might return a 500 currently if we're throwing an exception? Perhaps we could add a test for it?

db/migrations/versions/ba5e693514cd_.py
```diff
@@ -21,6 +21,7 @@ def upgrade():
     op.create_table(
         "assessment",
         sa.Column("id", sa.Text(), nullable=False),
+        sa.Column("compliance_status", sa.Text(length=36), nullable=False),
```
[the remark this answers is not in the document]
ramsharma-prog: I ran into some errors while running the migration. Trying to fix it

db/models/assessment.py
ramsharma-prog: Done

db/models/assessment.py
ramsharma-prog: Yeah, I agree. Deleted try-except

db/migrations/versions/ba5e693514cd_.py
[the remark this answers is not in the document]
ramsharma-prog: Ran migration again, all good now 😄

openapi/api.yml
gidsg: I think it should return a 404, as 401 is related to authorisation, can we make it return a 404 if assessment_id is not found?

ramsharma-prog: Yep, updated! 👍

gio-karunakaran (commented):
🎉 looks great! just a couple small comments, let me know if they make sense/ are valid! :D

gio-karunakaran (approved):

api/assessments/routes.py
```diff
@@ -52,3 +52,24 @@ def register(self):
...
+        search for an assessment_id & updates the status of
+        that assessment_id in the database.
+
+        Args: json data contains assessment_id & status
+
+        Returns: Updated status with assessment_id & application_id
+
+        """
+        json = request.get_json()
+        status = json.get("compliance_status")
```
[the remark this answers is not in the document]
ramsharma-prog: changed to `compliance_status`

db/models/assessment.py
```diff
@@ -11,13 +11,18 @@ class Assessment(db.Model):
...
     application_id = db.Column(db.Text(), index=True, unique=True)
 
     def __repr__(self):
         return f"<Assessment {self.id} for Application {self.application_id}>"
 
     def as_json(self):
-        return {"id": self.id, "applicationId": self.application_id}
+        return {
+            "id": self.id,
+            "status": self.compliance_status,
```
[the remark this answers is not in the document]
ramsharma-prog: yes, update to `compliance_status` 😄

PR #6, opened 2022-06-17
Fs 1009 standardise config

gio-karunakaran:
standardising configuration process in the assessment store repo


app.py
```diff
@@ -1,25 +1,38 @@
...
+    )
 
     flask_app = connexion_app.app
 
-    if testing:
+    if (os.environ.get("FLASK_ENV") == "development") | testing:
         flask_app.config.from_object(
             "config.environments.unit_testing.UnitTestingConfig"
         )
+        from config.environments.development import DevelopmentConfig
```
SentientHousePlant, line 30: 🔧 These configs do not match, should they?

SentientHousePlant (changes requested):

app.py
gio-karunakaran: 💯 that's sorted now and import is both from UnitTestingConfig 👍

harryyo (commented):
Just a couple of queries

requirements.txt
```diff
@@ -1,18 +1,62 @@
...
+pyparsing==3.0.9
+pyrsistent==0.18.1
+python-dotenv==0.20.0
+PyYAML==6.0
+requests==2.28.0
+rich==12.4.4
+ruamel.yaml==0.17.21
+ruamel.yaml.clib==0.2.6
 safety==1.10.3
-werkzeug==2.0.3
```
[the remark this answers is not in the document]
gidsg: We have similar problems in all the repos with random usage of `pip freeze`
The approach we took on digital marketplace was using `pip-compile` with a `requirements.in` file including the top level dependencies https://github.com/jazzband/pip-tools#example-usage-for-pip-compile I can spin off a card to adopt this approach if people like the sound of it?

app.py
```diff
@@ -1,25 +1,38 @@
...
+
 import connexion
 from config.env import env
 from connexion.resolver import MethodViewResolver
 from flask import Flask
 from openapi.utils import get_bundled_specs
+from utils.definitions import get_project_root
+
+
+project_root_path = str(get_project_root())
```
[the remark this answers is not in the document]
gio-karunakaran: 💯 I've removed this from this repo as we're not using it, but in other repos e.g. fund-store we are, but not 100% sure on the advantages of doing it this way vs using FLASK_ROOT variable? @SentientHousePlant will be better informed on this I think 😅

requirements.txt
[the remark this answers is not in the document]
gio-karunakaran: I'm happy with this 👍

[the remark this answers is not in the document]
harryyo: sounds good

config/environments/development.py
```diff
@@ -3,8 +3,10 @@
 from os import path
 
 from config.environments.default import Config
+from fsd_tech import configclass
 
 
+@configclass
```
Adam-W1, line 9: python Q - if our env config (dev.py, prod.py etc) inherit from default.py env (which already has the @configclass decorator on the base Config class), do we also need to apply this decorator to all sub classes ?

app.py
```diff
@@ -19,13 +17,10 @@ def create_app(testing=False) -> Flask:
...
     flask_app = connexion_app.app
 
-    if (os.environ.get("FLASK_ENV") == "development") | testing:
+    if testing:
         flask_app.config.from_object(
             "config.environments.unit_testing.UnitTestingConfig"
         )
-        from config.environments.unit_testing import UnitTestingConfig
-
-        UnitTestingConfig.pretty_print()
```
SentientHousePlant: 🔧 I think this can be changed so that we set testing in pytest.ini instead and the corresponding config is loaded.

config/__init__.py
```diff
@@ -2,13 +2,23 @@
...
+            DevelopmentConfig as Config,
+        )  # noqa
+    case "dev":
+        pass
+    case "test":
+        from config.environments.test import TestConfig as Config  # noqa
+    case "production":
+        pass
+    case _:
+        from config.environments.default import DefaultConfig as Config  # noqa
```
SentientHousePlant, line 17: 🔧 Add unit_test and dev a la discovery frontend.

config/environments/default.py
```diff
@@ -1,22 +1,20 @@
 """Flask configuration."""
 from os import environ
-from os import path
+from pathlib import Path
 
 from fsd_tech import configclass
```
SentientHousePlant, line 5: 🔧 Should be fsd_utils

tests/mocks/sqlite_test_db.py
```diff
@@ -1,14 +1,14 @@
...
+from config.environments.unit_testing import UnitTestingConfig
 
 
 class SqliteTestDB:
     @classmethod
     def remove(cls):
-        flask_root = env.config.get("FLASK_ROOT")
-        db_file_name = env.config.get("SQLITE_DB_NAME")
+        flask_root = UnitTestingConfig.FLASK_ROOT
+        db_file_name = UnitTestingConfig.SQLITE_DB_NAME
```
SentientHousePlant, line 11: 🔧 Update to pull from config.Config.

SentientHousePlant (changes requested):
Some changes to make, mainly to update it to the freshest config standards.

app.py
gio-karunakaran: 💯 have included the latest config set-up via pytest.ini 👍

config/__init__.py
gio-karunakaran: 👍

config/environments/default.py
gio-karunakaran: 💯 done and dusted!

tests/mocks/sqlite_test_db.py
gio-karunakaran: 👍

config/__init__.py
```diff
@@ -1,14 +1,31 @@
...
-    from config.environments.test import TestConfig as Config
-elif FLASK_ENV == "production":
-    from config.environments.production import ProductionConfig as Config
-else:
-    from config.environments.default import Config  # noqa
+match FLASK_ENV:
+    case "development":
+        from config.environments.development import (
+            DevelopmentConfig as Config,
+        )  # noqa
```
SentientHousePlant, line 10: ⛏️ You can now remove these comments

SentientHousePlant (changes requested):

SentientHousePlant (approved):
Great stuff Gio 🤖

config/__init__.py
```diff
@@ -1,14 +1,31 @@
...
+        from config.environments.unit_testing import (
+            UnitTestingConfig as Config,
+        )
+    case "production":
+        from config.environments.production import (
+            ProductionConfig as Config,
+        )
+    case _:
+        from config.environments.default import DefaultConfig as Config  # noqa
+
```
ramsharma-prog, line 25: nice use of Pattern Matching 💯

ramsharma-prog:
Great work Gio 💯

config/__init__.py
```diff
@@ -1,14 +1,31 @@
...
-elif FLASK_ENV == "production":
-    from config.environments.production import ProductionConfig as Config
-else:
-    from config.environments.default import Config  # noqa
+match FLASK_ENV:
+    case "development":
+        from config.environments.development import (
+            DevelopmentConfig as Config,
+        )
+    case "dev":
```
Adam-W1, line 11: I think you need to call in config.environments.dev here

gio-karunakaran: 💯 all sorted!

SentientHousePlant (approved):

SentientHousePlant (approved):

pytest.ini
```diff
@@ -0,0 +1,23 @@
+[pytest]
+env =
+    FLASK_ENV=unit_test
+    FLASK_DEBUG=1
```
Adam-W1, line 4: @SentientHousePlant @gio-karunakaran . Do we add 'FORCE_HTTPS = False' in unit-test config (config/environments/unit_test.py) rather than set it in this file? I was forced to do it that way in authenticator because we check that the flask app is not in debug mode when testing and setting that ^ forces the app into debug mode which we cannot overwrite afaik.

Adam-W1 (commented):
Just a few comments - great work Gio 🦸‍♂️

pytest.ini
harryyo: I don't know if the debug mode test is particularly important? Do other tests fail if it is in debug mode? Could we remove it? - Equally perhaps is it required for any tests to have debug mode on? i.e. do we care either way?

requirements.txt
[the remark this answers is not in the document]
harryyo: This still seems to be the full pip freeze - are we agreed it would be worth sliming this down (with whatever package we use)? If we don't do it now in this ticket, it might be worth making a ticket to add pip-compile to repos.

harryyo (commented):
Couple of comments. Also test_sqlite.db should not be committed (so should be removed). When pytest runs it should be deleted automatically when the tests complete. Can you check that it is getting removed?

requirements.txt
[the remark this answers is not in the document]
gio-karunakaran: have pip compiled this now 👍 happy for us to add a ticket on the board to get this consistent across the repos 💯

requirements.txt
```diff
@@ -1,18 +1,71 @@
...
+openapi-schema-validator==0.2.3
+openapi-spec-validator==0.4.0
+ordered-set==4.1.0
+packaging==21.3
+pathspec==0.9.0
+pbr==5.9.0
+platformdirs==2.5.2
+pluggy==1.0.0
+prance==0.21.8.0
+pre-commit==2.19.0
```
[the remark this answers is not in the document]
gio-karunakaran: all sorted now! 💯

pytest.ini
gio-karunakaran: @Adam-W1 I had also set 'FORCE_HTTPS = False' in unit-test config and have now removed FLAS_DEBUG=1 in this file and tests are still running fine so we can stick with this? 👍

gio-karunakaran: @harryyo ^

dev-requirements.txt
```diff
@@ -0,0 +1,257 @@
+#
+# This file is autogenerated by pip-compile with python 3.10
+# To update, run:
+#
+#    pip-compile dev-requirements.in
```
harryyo, line 5: Could you briefly update the README so it's clear that this is how requirements are installed now.

.github/workflows/deploy.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         with:
           python-version: 3.10.1
       - name: install dependencies
-        run: python -m pip install --upgrade pip && python -m pip install -r requirements.txt
+        run: python -m pip install --upgrade pip && python -m pip install -r dev-requirements.txt
```
harryyo, line 26: Should this just be requirements.txt?

.github/workflows/deploy.yml
```diff
@@ -73,9 +73,9 @@ jobs:
         with:
           python-version: 3.10.1
       - name: install dependencies
-        run: python -m pip install --upgrade pip && python -m pip install -r requirements.txt
+        run: python -m pip install --upgrade pip && python -m pip install bandit safety
       - name: Bandit
-        run: bandit -r ./auth ./demo
+        run: bandit -r . -x config/*,tests/*
```
harryyo: What's the update doing here?

harryyo (commented):
Couple of questions

.github/workflows/deploy.yml
gio-karunakaran: we're getting bandit to run on all files excluding (-x) config and test folders 👍

dev-requirements.txt
gio-karunakaran: 💯 done!

.github/workflows/deploy.yml
harryyo: ah of course, sorry

harryyo (approved):
Great

PR #7, opened 2022-06-28
FS-1112 Implement data model in Assessment for criteria and Subcriteria

gio-karunakaran:
adding the data model for subcriterias

### DIFF
--- a/db/models/sub_criteria.py
+++ b/db/models/sub_criteria.py
@@ -1,0 +1,66 @@
+import uuid
+
+from db import db
+
+
+class SubCriteria(db.Model):
+    sub_criteria_id = db.Column(
+        "sub_criteria_id",
+        db.Text(),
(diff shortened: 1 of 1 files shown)


db/models/sub_criteria.py
```diff
@@ -0,0 +1,66 @@
+import uuid
+
+from db import db
+
+
+class SubCriteria(db.Model):
+    sub_criteria_id = db.Column(
+        "sub_criteria_id",
+        db.Text(),
+        default=lambda: str(uuid.uuid4()),
```
Adam-W1, line 10: can this be type uuid as oppose to a string?

gio-karunakaran: great spot! have removed the str(), uuid has an appropriate __repr__ 👍

Adam-W1 (approved):

PR #9, opened 2022-07-04
FS-840-Part-A-Scores-And-Justifications-Table

gio-karunakaran:
adding table to record scores and justifications


gidsg (changes requested):
We need to create this using a migration in order for it to work on Test/on Postgres database.

e.g. run `flask db migrate -m "add scores justifications table"`

db/models/scores_justifications.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, message="Sorry, there was a problem, please try later"):
+        self.message = message
+        super().__init__(self.message)
+
+
+class ScoresJustificationsMethods:
+    @staticmethod
+    def score_justification(as_json=False):
+        scores_justifications = ScoresJustifications.query.all()
```
gidsg, line 66: Is it actually useful to get all scores? Would we not want scores by assessment ID or something?

db/models/scores_justifications.py
```diff
@@ -0,0 +1,80 @@
...
+                for Criteria {self.criteria_id} \
+                of Round {self.round_id}>"
+
+    def as_json(self):
+        return {
+            "sub_criteria_id": self.sub_criteria_id,
+            "round_id": self.round_id,
+            "criteria_id": self.criteria_id,
+            "sub_criteria_title": self.sub_criteria_title,
+        }
```
gidsg, line 48: Can we include the other fields like `assessment_id`, `score` and `justification` to make this more useful?

db/models/scores_justifications.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class ScoresJustifications(db.Model):
+    scores_justifications_id = db.Column(
+        "scores_justifications_id",
+        db.Text(),
+        default=lambda: uuid.uuid4(),
+        primary_key=True,
+    )
+    timestamp = db.Column("timestamp", db.Text(), default=str(datetime.now()))
```
gidsg, line 14: Can we use https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.DateTime for this? And default to `datetime.utcnow` and call it `created_at`?

db/models/scores_justifications.py
```diff
@@ -0,0 +1,80 @@
...
+from db import db
+
+
+class ScoresJustifications(db.Model):
+    scores_justifications_id = db.Column(
+        "scores_justifications_id",
+        db.Text(),
+        default=lambda: uuid.uuid4(),
+        primary_key=True,
+    )
```
gidsg, line 13: Can we use `UUIDType` here instead? This will use the Postgres UUID native type
e.g.
```
id = sa.Column(
                UUIDType(binary=False),
                primary_key=True,
                default=uuid.uuid4
            )
```

SentientHousePlant: I agree with this

db/models/scores_justifications.py
```diff
@@ -0,0 +1,80 @@
...
+        db.Integer(),
+    )
+    justification = db.Column(
+        db.Text(),
+    )
+
+    def __repr__(self):
+        return f"<Sub-Criteria {self.sub_criteria_title}, {self.sub_criteria_id} \
+                for Criteria {self.criteria_id} \
+                of Round {self.round_id}>"
```
SentientHousePlant, line 40: ⛏️ Reprs should resemble how the object is created ie `ScoresJustification(score=5,blah=blah)`, but this would make a great `__str__`

SentientHousePlant (changes requested):

PR #11, opened 2022-07-06
Fs 840 part c form hook up

gio-karunakaran:
adding scores and justifcations models, methods and tests


api/scores_justifications/routes.py
```diff
@@ -0,0 +1,51 @@
+from api.responses import error_response
+from api.responses import scores_justifications_response
+from api.responses import scores_justifications_response_list
+from db.models.scores_justifications import ScoresJustificationsError
+from db.models.scores_justifications import ScoresJustificationsMethods
+from flask.views import MethodView
+
+
+class ScoresJustificationsView(ScoresJustificationsMethods, MethodView):
```
Adam-W1, line 9: Im just pondering directory/file structure here. I understand it mirrors our db model classes and is called routes.py because our api.yml routes are defined here, but would it make more sense for these files be inside a routes folder within api rather than api directory filled with directories based on the different database models, which will be different for each repo. If im missing the point please could we have a call :)

tests/mocks/sqlite_test_db.py
```diff
@@ -17,3 +23,43 @@ def remove(cls):
...
+            sub_criteria_title="nothing",
+        )
+        db.session.add(sub_criteria_2)
+        db.session.commit()
+
+        score_justification = ScoresJustifications(
+            scores_justifications_id=uuid.UUID(
+                "123e4567-e89b-12d3-a456-426655440003"
+            ),
+            created_at=dateutil.parser.parse("2022-07-07T09:11:38.240578Z"),
```
gidsg, line 57: There is no need for this library, as it appears we're just using it to parse a string into a date. We can do this with the core library e.g.
`datetime.fromisoformat('2022-07-07')`

api/scores_justifications/routes.py
gio-karunakaran: that is a very good point! was thinking we should have a chat about this also and agree on the best way of doing this 👍🏾

tests/mocks/sqlite_test_db.py
gio-karunakaran: added strptime from datetime module to resolve this 👍🏾

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
gidsg, line 1: Can you revert this file, guess it was committed by accident.

db/migrations/versions/592143492996_add_scores_justifications_table.py
```diff
@@ -0,0 +1,58 @@
...
+        sa.PrimaryKeyConstraint("sub_criteria_id"),
+    )
+    op.create_table(
+        "scores_justifications",
+        sa.Column(
+            "scores_justifications_id", UUIDType(binary=False), nullable=False
+        ),
+        sa.Column("created_at", sa.DateTime(), nullable=True),
+        sa.Column("assessment_id", sa.Text(), nullable=True),
+        sa.Column("person_id", sa.Text(), nullable=True),
```
gidsg, line 37: Can we call this user ID?

db/migrations/versions/592143492996_add_scores_justifications_table.py
```diff
@@ -0,0 +1,58 @@
...
+        sa.Column("sub_criteria_id", sa.Text(), nullable=True),
```
gidsg, line 38: For these short text fields can we use `String` type with a limit of 255 length? Else we will be storing them inefficiently in the database.

.vscode/launch.json
gio-karunakaran: reverted! sorry about that!

db/migrations/versions/592143492996_add_scores_justifications_table.py
gio-karunakaran: 👍🏾

db/migrations/versions/ba5e693514cd_.py
```diff
@@ -1,42 +0,0 @@
-"""empty message
```
gidsg, line 1: Looks like the previous migrations have been accidentally deleted. Please can we restore them?

db/migrations/versions/cd404dd517c5_.py
```diff
@@ -0,0 +1,59 @@
...
+down_revision = "4dd55a3dc51e"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "sub_criteria",
+        sa.Column("sub_criteria_id", sa.Text(), nullable=False),
```
gidsg, line 23: Should we use UUID for this too? Also, can we just call these fields `id` as its implied by the table name rather than `sub_criteria_id`?

gio-karunakaran: have changed this and updated methods and tests to reflect this 👍🏾

api/routes/subcriterias/routes.py
```diff
@@ -14,9 +15,11 @@ def list(self, assessment_id: str = None):
         GET /assessments/{assessment_id}/sub_criterias endpoint
         :return: Json Response
         """
+        print("hello")
```
SentientHousePlant, line 18: 🚨 print!

SentientHousePlant (approved):
Approved but check out my print comment

api/routes/subcriterias/routes.py
gio-karunakaran: deleted!

gidsg (approved):
LGTM 👍

SentientHousePlant (changes requested):
delete test_sqlite.db

SentientHousePlant (approved):
Wow! A massive ticket, that's impressive. Well done for getting this done :)

PR #16, opened 2022-07-22
Fs 722 comments

ramsharma-prog:


api/routes/scores_justifications/routes.py
```diff
@@ -1,8 +1,10 @@
+import sqlalchemy
```
gidsg, line 1: Is this import unused?

config/envs/default.py
```diff
@@ -32,9 +32,10 @@ class DefaultConfig:
     APPLICATION_ENDPOINT = "/applications/{account_id}"
 
     # Fund Store
-    FUND_STORE_API_HOST = environ.get("FUND_STORE_API_HOST")
+    FUND_STORE_API_HOST = environ.get("FUND_STORE_API_HOST", "fund_score_host")
```
gidsg, line 35:
```suggestion
    FUND_STORE_API_HOST = environ.get("FUND_STORE_API_HOST", "fund_store_host")
```
Doesn't seem like this fallback is used at all so can it just be removed?

gidsg:
Need to fix the conflicts here and rebase so that database migrations are sequential

PR #39, opened 2022-10-21
Fs 1669 create and store comments

Adam-W1:
Create and store comments in the assessment store


db/models/comments.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+    sub_criteria_id = db.Column(
+        UUIDType(binary=False),
+        db.ForeignKey(SubCriteria.id),
+        nullable=False,
+    )
+    section_id = db.Column(
+        db.Integer(),
+        nullable=False
+    )
```
harryyo, line 40: Appreciate this wasn't described as a requirement in the ticket - but would it be worth considering making this a UUID with a foreign key to an actual Sections model? It feels a bit odd making a comment model without making a model for the object that is being commented? Or is there a ticket for making the sections model?

db/models/comments.py
```diff
@@ -0,0 +1,138 @@
...
+        comment = Comment.query.get(id)
+        if not comment:
+            raise CommentError(
+                message="Comment could not be found"
+            )
+        return comment
+
+    @staticmethod
+    def get_section_comments(
+        assessment_id: str, sub_criteria_id: str, section_id: int, as_json=False
```
harryyo, line 92: Is there a reason that section_id needs to be an int? If not, maybe it should be a str in common with other ids for objects?

db/models/comments.py
```diff
@@ -0,0 +1,138 @@
...
+            db.session.rollback()
+            message = e.orig.args[0] if e.orig else f"There was a StatementError following the database query: {e.statment}"
+            raise CommentError(message)
+        return comments
+
+    @staticmethod
+    def save_comment(
+        assessment_id: str,
+        sub_criteria_id: str,
+        section_id: int,
```
harryyo, line 116: Maybe str?

harryyo (commented):
Looks great - only comments are about whether section_id should be a str/uuid. and potentially a section model unless there is a ticket for creating a section model?

db/models/comments.py
Adam-W1: yes, I think this will be the next iteration after our talk on the db structure today. Originally we had the idea that a section would be an INT as per convos with @stevewalton28 on mural [here ](https://app.mural.co/t/mhclg2837/m/mhclg2837/1666017408703/070f1702ce92e818d80fd6af184a258f8ed06768?sender=u7723e74e80c664f54eda3069) but as part of https://digital.dclg.gov.uk/jira/browse/FS-1671 we will be changing this to a separate table with an uuid link as you say 👍

harryyo (approved):

PR #41, opened 2022-11-13
FS-1897 (Part A) Add data model.

SentientHousePlant:
# Things to note.
Lots of the old tables have been removed in this PR. I'll create a branch to hold these and we can add them back in as we get a clearer picture of the data model.

# Overview.
- Added AssessmentRecords.
- Added some useful queries.
- Added some tests.
- Added lots of useful pytest stuff.
- Marshmallow is used for serialisation over writting custom "to_dict" functions. 
- Better seperation of concerns. Functionality related to querying data and testing is now fully seperated from the table definitions.

# Things I need to do still.
- [x] Add docstrings.
- [x] Add README.md for each folder. 

Contrary to the name of the ticket I did not add application importing, but all the inserting functions are there ready for me to write the ~10 lines to download stuff from the app store.


db/models/assessment_record/assessment_records.py
```diff
@@ -0,0 +1,48 @@
...
+from sqlalchemy import func
+from sqlalchemy.dialects.postgresql import ENUM
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.dialects.postgresql import TEXT
+
+
+class AssessmentRecords(db.Model):
+    __tablename__ = "assessment_records"
+
+    application_id = db.Column(
```
Adam-W1, line 14: is there an arguemnt for this being uuid and the short-name being more hu8man readbale ?

.github/workflows/deploy.yml
```diff
@@ -38,6 +38,15 @@ jobs:
...
           command: add-network-policy funding-service-design-assessment-store-dev funding-service-design-notification-dev --protocol tcp --port 8080
+    - name: Apply network policy for application store service for dev
+      uses: citizen-of-planet-earth/cf-cli-action@v2
+      with:
+          cf_api:      ${{secrets.CF_API}}
+          cf_org:      ${{secrets.CF_ORG}}
+          cf_space:    ${{secrets.CF_SPACE }}
+          cf_username: ${{secrets.CF_USER}}
+          cf_password: ${{secrets.CF_PASSWORD}}
+          command: add-network-policy funding-service-design-assessment-store-dev funding-service-design-application-store-dev --protocol tcp --port 8080
```
gidsg, line 49: Are we using this network policy? Doesn't seem like any reference to application-store in the code?

SentientHousePlant: In preperation for when we query the app store for the jsons.

config/envs/development.py
```diff
@@ -22,3 +22,9 @@ class DevelopmentConfig(DefaultConfig):
 
     # Database
     SQLALCHEMY_TRACK_MODIFICATIONS = False
+
+    DEV_SQLALCHEMY_DATABASE_URI = (
+        DefaultConfig.SQLALCHEMY_DATABASE_URI + "_DEV"
```
gidsg, line 27: Why are we changing this value? Don't think this is is consistent with what we do for other apps?

SentientHousePlant: I changed it so that using an invoke task would create a specially named db, although on reflection that can be set in the invoke task so that config isnt changed.

db/migrations/versions/03b7d7a4e1c3_.py
```diff
@@ -0,0 +1,78 @@
...
+        ),
+    )
+    op.create_entity(public_assessment_records_block_updates_on_app_blob)
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    public_assessment_records_block_updates_on_app_blob = PGTrigger(
```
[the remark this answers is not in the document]
SentientHousePlant: I think it uses the signiture to key the drop command.

.github/workflows/deploy.yml
SentientHousePlant: But will move it out for now and include in the PR where that functionality is added, if that is agreeable.

db/models/assessment_record/db_extras.py
```diff
@@ -0,0 +1,43 @@
...
+block_json_updates_trig = PGTrigger(
+    schema="public",
+    signature="block_updates_on_app_blob",
+    definition="""
+    BEFORE UPDATE
+    ON assessment_records
+    FOR EACH ROW
+    EXECUTE PROCEDURE block_blob_mutate();""",
+    on_entity="assessment_records",
+)
```
Adam-W1, line 43: lovely use of the trigger to stop us doing things we should not do here

Adam-W1: maybe we just call this file db_triggers?

tests/helpers.py
```diff
@@ -1,89 +1,45 @@
...
-
-    diff = DeepDiff(
-        expected_data,
-        response_data,
-        exclude_paths=exclude_paths,
-        exclude_regex_paths=exclude_regex_paths,
-    )
+@contextmanager
+def no_gather_sql():
+    @event.listens_for(Engine, "before_cursor_execute", retval=True)
```
Adam-W1, line 12: whats the purpose of this?

tests/helpers.py
```diff
@@ -1,89 +1,45 @@
...
+    return (
+        db.session.query(table)
+        .offset(
+            func.floor(
+                func.random()
+                * db.session.query(func.count(primary_key_column))
+            )
+        )
+        .limit(1)
+        .one()
```
Adam-W1, line 44: does limit(1) and one() do the same thing?

tests/sql_infos.py
```diff
@@ -0,0 +1,145 @@
...
+
+            time_string.append(time)
+
+            during = Text(f"{query['during']}", style="italic black")
+            during_string.append(during)
+
+            fancy_print(statement_string)
+            fancy_print(statement)
+            fancy_print(time_string)
+            fancy_print(during_string)
```
Adam-W1, line 145: lovely feedback

db/models/assessment_record/assessment_records.py
SentientHousePlant: Changed to uuid.

tests/application_store_json.py
```diff
@@ -0,0 +1,921 @@
+# flake8: noqa
+from string import Template
+
+application_store_json_template = Template(
```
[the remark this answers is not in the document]
SentientHousePlant: Welsh ones will be translated prior to loading in.

tests/helpers.py
SentientHousePlant: limit refers to the sql query itself. It adds the LIMIT clause to the query. It gurentees that only a single result will be returned when pg executes the query.

One is a python side validator which will confirm that only one result was returned and extract it for us to use. 

Removing limit(1) causes an exception to be raised.

tests/helpers.py
SentientHousePlant: We don't want to time sql queries which are executed as test helpers, because they don't reflect sql which will run in production.

SentientHousePlant: This context manager adds a comment to those queries which makes the timer ignore them.

tests/db_seed_data.py
```diff
@@ -0,0 +1,79 @@
...
+
+    places = [
+        "skatepark",
+        "pub",
+        "cinema",
+        "community centre",
+    ]
+
+    cities = [
+        "Bath",
```
gidsg, line 37: Can we use fixed data rather than random generation? (So we can rely on these records for testing known values.)

srh-sloan (commented):
Really like that the unit tests print out the SQL as they run. But could we not use random values for unit tests please? Makes it hard to debug and can lead to inconsistent tests, like if it fails on one row and then when you rerun it it gets a different row it would pass - harder to debug.

Also can you update the main README file please as it still mentions running `invoke bootstrap-test-db`

db/migrations/versions/d7eed6243143_.py
```diff
@@ -1,151 +0,0 @@
...
-# revision identifiers, used by Alembic.
-revision = "d7eed6243143"
-down_revision = None
-branch_labels = None
-depends_on = None
-
-
-def upgrade():
-    # ### commands auto generated by Alembic - please adjust! ###
-    op.create_table(
```
gidsg, line 22: If we are deleting migrations, I guess we'll need to re-create the database on test?

db/models/assessment_record/assessment_records.py
```diff
@@ -0,0 +1,82 @@
...
+        "application_id", UUID(), primary_key=True
+    )
+
+    short_id = db.Column("short_id", db.Text(), nullable=False)
```
gidsg, line 20: Is this meant to be the friendly application reference, if so this could be a String type rather than text,

db/models/assessment_record/assessment_records.py
```diff
@@ -0,0 +1,82 @@
...
+
+    type_of_application = db.Column(
+        "type_of_application", db.Text(), index=True, nullable=False
```
gidsg, line 23: Can we make this a string type rather than text?

db/models/assessment_record/assessment_records.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+
+    project_name = db.Column(
+        "project_name", db.Text(), index=True, nullable=False
```
gidsg, line 27: Can we make this a string type rather than text?

db/models/assessment_record/assessment_records.py
```diff
@@ -0,0 +1,82 @@
...
+
+    funding_amount_requested = db.Column(
+        "funding_amount_requested", db.Float(), index=True, nullable=False
+    )
+
+    round_id = db.Column("round_id", UUID(), index=True, nullable=False)
+
+    fund_id = db.Column("fund_id", UUID(), index=True, nullable=False)
+
+    langauge = db.Column("langauge", db.Text(), default="en", nullable=False)
```
gidsg, line 38: Can we make this a String type rather than Text? (Or an ENUM with 'cy' and 'en' as values?

db/migrations/versions/d7eed6243143_.py
SentientHousePlant: Yes, because the whole schema has changed since we last merged to main.

db/queries/assessment_records/helpers.py
```diff
@@ -0,0 +1,48 @@
...
+def get_mapper(application_type):
+
+    match application_type:
+
+        case "COF":
+            return COF_json_mapper
+
+        case "citizan_space":
+            return {"id": "some key in some json somewhere."}
+
```
gidsg, line 28: Can we remove speculative changes like this?

db/models/assessment_record/assessment_records.py
SentientHousePlant: How about the postgres specific TEXT since it's a infinite varchar by default like String?

db/queries/assessment_records/helpers.py
```diff
@@ -0,0 +1,48 @@
...
+COF_json_mapper = {
+    # "column name" : "jsonpath to value in jsonb blob".
+    "application_id": "$.id",
+    "project_name": "$.project_name",
+    "short_id": "$.reference",
+    "fund_id": "$.fund_id",
+    "round_id": "$.round_id",
+    "funding_amount_requested": (
+        '$.forms[*].questions[*].fields[*] ? (@.key == "JzWvhj")."answer"'
+    ),
```
gidsg, line 15: Not sure if this is better pulled through to the top level in an import script?

Also I think the intention is for this to be capital + revenue  so `JzWvhj` + `jLIgoi`

db/queries/assessment_records/queries.py
```diff
@@ -0,0 +1,74 @@
...
+        row = {**derived_values, "jsonb_blob" : loaded_json, "type_of_application" : application_type}
+
+        rows.append(row)
+
+        del loaded_json
+
+    db.session.bulk_insert_mappings(AssessmentRecord, rows)
+    db.session.commit()
+
+def find_answer_by_key_runner(field_key: str, app_id: str):
```
gidsg, line 61: What's the thinking behind this being called `runner` could it be just `find_answer_by_key`?

pytest.ini
```diff
@@ -3,7 +3,7 @@ env =
     FLASK_ENV=unit_test
     FLASK_DEBUG=1
     FUND_STORE_API_HOST=https://funding-service-design-fund-store-test.london.cloudapps.digital
-    D:DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:5432/fsd_assess_store_test
+    D:DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:5432/fsd_assess_store
```
gidsg, line 6: When is this used? Shouldn't it match one of the database names defined in the config?

tasks.py
```diff
@@ -1,16 +1,142 @@
...
+    echo_print("Profiling tests.")
+    c.run(
+        "python -m cProfile -s tottime -m pytest"
+        f" > {resultsfile}"
+    )
+    echo_print(f"Results saved in f{resultsfile}")
+
+
+@task
+def bootstrap_dev_db(c):
```
gidsg, line 59: Are these tasks documented somewhere? When are they intended to be used? Do they need adding to the readme?

db/models/assessment_record/assessment_records.py
gidsg: I thought String mapped to varchar ref: https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.String so not infinite, and better for this kind of case (short strings)?

I think Text would be fine for things like comment body but String for fields like this.

adamdavies1 (approved):
:+1:

db/queries/assessment_records/helpers.py
SentientHousePlant: Thank you! 😃

tasks.py
SentientHousePlant: Writing up READMEs now

pytest.ini
SentientHousePlant: Good point! I think this must be an artefact of before we had the config management we have now. Will remove!

db/queries/assessment_records/queries.py
SentientHousePlant: This specifically uses the runners schema to find questions, however since we are currently only using runner jsons I can rename.

db/models/assessment_record/assessment_records.py
SentientHousePlant: Ok, I will swap for a String(256)!

SentientHousePlant: (255 I mean)

gio-karunakaran (approved):
LGTM 👍🏾 amazing stuff!!

PR #42, opened 2022-11-16
Mapping configuration for assessment

Adam-W1:
Map answered questions to assessment sub criteria and sections

https://digital.dclg.gov.uk/jira/browse/FS-1829

structure of config:
```
- cofr2w2 contains (ordered):
 - unscored sections contains [{sub criteria}] contains [{themes}] contains [{answers}]
 - scored criteria contains [{sub criteria}] contains [{themes}] contains [{answers}]
```
An answer dictionary contains 
```
{
    "field_id": ", // used to extract an answer from the matching field_id in the application payload
    "form_name": ", // used to help traverse the applciation payload (unless flattened in which case this can be removed)
    "field_type": "",  // field type at source (in the form json)
    "presentation_type": "", // field type used by assessment service to determine template component useage
    "question": "", // used to provide context - will this be used instead of the question provided on the applciation json?
},
```


SentientHousePlant (commented):
I like this! Makes sense that themes are children of sub criteria. 

If you want to be really extra you could write a schema validator à la https://github.com/keleshev/schema or https://github.com/python-jsonschema/jsonschema so that future schemas conform to a validatable standard.

Adam-W1:
> Some additional food for thought : you could define some structures outside of the hierarchy and use an internal reference to them. This is what we do in the open api spec with components.
> 
> This would allow for some modularity, which might be useful as this structure expands.
> 
> See https://pypi.org/project/jsonref/ for an example of this.

love those suggestions ❤️

mapping.json
```diff
@@ -0,0 +1,111 @@
...
+        "name": "Organisation information",
+        "sub_criteria": [{
+            "id": "org_info",
+            "name": "Organisation information",
+            "themes": [{
+                "id": "general_info",
+                "name": "General information",
+                "answers": [{
+                    "field_id": "pipopl",
+                    "field_type": "textarea"
```
gidsg, line 14: Could this be derived from the `type` field in the submitted JSON? So maybe we don't need it here as well?

mapping.json
```diff
@@ -0,0 +1,111 @@
+{
+    "schema_id": "cof_r2_assessment",
```
srh-sloan, line 2: should we call it `cof_r2w2_assessment` as it maps directly then to the short names of fund/round

mapping.json
Adam-W1: I had this discussion with @stevewalton28 and @harryyo yesterday. 
The reason we kept it in is to have a simple toggle when deciding how to render the field, the runner might say it's an XX type, but we may want to display it in a text field template etc. If I have understood correctly, It gives more responsibility to the assessment side in presentation

mapping.json
```diff
@@ -0,0 +1,111 @@
...
+        "id": "org_info",
+        "name": "Organisation information",
+        "sub_criteria": [{
+            "id": "org_info",
+            "name": "Organisation information",
+            "themes": [{
+                "id": "general_info",
+                "name": "General information",
+                "answers": [{
+                    "field_id": "pipopl",
```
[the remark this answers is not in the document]
Adam-W1: Nice suggestion 💯

mapping.json
Adam-W1: sounds sensible :)

config/mappings/test_mapping_config.py
```diff
@@ -0,0 +1,13 @@
...
+
+with open("config/mappings/mapping.json", "r") as mapping_file:
+    mapping = json.load(mapping_file)
+
+try:
+    validate(instance=mapping, schema=schema)
+except Exception:
+    raise Exception
+finally:
+    print("Mapping conforms with schema.")
```
SentientHousePlant, line 13: Validate does this for us!

In the future there will be many mapping files. So you could loop through these files using:
```python
for file in os.listdir(directory):
   mapping = json.load(mapping_file)
   validate(instance=mapping, schema=schema)
# Runs if nothing weird happens in for loop.
else:
   print("All mappings are valid!")
```

```suggestion
validate(instance=mapping, schema=schema)
print("Mapping conforms with schema.")
```

srh-sloan, line 13: Doesn't finally run even if validation fails? So I think you need to move the `print("Mapping conforms with schema.")` inside the `try:` clause

mapping.json
gidsg: OK makes sense.

.flaskenv
```diff
@@ -2,3 +2,4 @@ FLASK_APP=app.py
 FLASK_ENV=development
 FLASK_RUN_PORT=5000
 FLASK_RUN_HOST=localhost
+DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:5432/assessment_spike
```
gidsg, line 5: Not sure we want to commit this?

Adam-W1: My bad 😱

srh-sloan (approved):
Looks good, I like that it's split into a few smaller files now too 👍

config/mappings/test_mapping_config.py
```diff
@@ -0,0 +1,7 @@
+import json
+from jsonschema import validate
+from assessment_mapping_schema import top_level_assessment_mapping_schema as schema
+from assessment_mapping import cof_r2w2_assessment as mapping
+
+validate(instance=mapping, schema=schema)
+print("Mapping conforms with schema.")
```
tferns, line 7: This could maybe be added to the tests that are ran by `pytest`?

Adam-W1: Good call, ill hold off for now as this might be moving to assessment

PR #45, opened 2022-11-24
FS-1581: Add endpoint to consume mapping configuration

tferns:
Tentative PR, don't branch on top of this.


db/queries/assessment_records/queries.py
```diff
@@ -105,3 +105,26 @@ def find_answer_by_key_runner(field_key: str, app_id: str) -> List[tuple]:
...
+    :type application_id: str
+    :return: The matching row from the `assessment_records` table.
+    :rtype: AssessmentRecord
+    """
+
+    assessment_record = (
+        db.session.query(AssessmentRecord)
+        .filter(AssessmentRecord.application_id == application_id)
+        .one()
+    )
```
SentientHousePlant, line 124: ⛏️ Change to new SQLA syntax. So, `stmt = select(AssessmentRecord).where(AssessmentRecord.application_id == application_id)` and then execute the stmt using `db.session.scalar(stmt).one()`.

db/schemas/schemas.py
```diff
@@ -15,3 +15,14 @@ class Meta:
...
+
+class AssessmentRecordMetadataWithJson(SQLAlchemyAutoSchema):
+    """AssessmentRecordMetadata The marshmallow class used to turn SQLAlchemy
+    rows into json for return in http responses."""
+
+    class Meta:
+        model = AssessmentRecord
+
+    workflow_status = Enum(Status)
+    language = Enum(Language)
```
SentientHousePlant, line 28: ⛏️ Instead of this we can alter the existing Schema and remove the exclude. This will mean that the we will need to add some more `defer`s to the current queries.

openapi/assessment_config.yml
```diff
@@ -0,0 +1,79 @@
...
+          type: string
+        name:
+          type: string
+        weighting:
+          type: number
+          format: float
+        sub_criteria:
+          type: array
+          items:
+            $ref: '#/components/schemas/SubCriteria'
```
SentientHousePlant, line 66: ⛏️ No need for both of these, if you pick the bigger schema you can just use that. The most important thing is that `scored_criteria` **requires** a weighting. See https://github.com/communitiesuk/funding-service-design-notification/blob/main/openapi/components.yml#L147 for how to inherit from a schema and make a property required.

tasks/db_tasks.py
```diff
@@ -84,7 +84,7 @@ def seed_dev_db(c):
                 f" {apps * rounds * funds} test rows."
             )
 
-            seed_database(apps, rounds, funds)
+            seed_database_randomly(apps, rounds, funds)
```
SentientHousePlant, line 87: 💯 Thank you!

db/queries/assessment_records/queries.py
```diff
@@ -105,3 +105,26 @@ def find_answer_by_key_runner(field_key: str, app_id: str) -> List[tuple]:
...
+
+
+def find_assessment(application_id: str):
+    """find_assessment Given an application id `application_id` we return the
+    matching row from the `assessment_records` table.
+
+    :param application_id: The application id of the queried row.
+    :type application_id: str
+    :return: The matching row from the `assessment_records` table.
+    :rtype: AssessmentRecord
```
SentientHousePlant, line 117: 💯 Lovely Docstring!

db/queries/assessment_records/queries.py
```diff
@@ -105,3 +105,26 @@ def find_answer_by_key_runner(field_key: str, app_id: str) -> List[tuple]:
         .filter(AssessmentRecord.application_id == app_id)
         .one()
     )
+
+
+def find_assessment(application_id: str):
```
SentientHousePlant, line 110:
```suggestion
def find_assessment(application_id: str) -> Dict:
```

config/mappings/test_mapping_config.py
```diff
@@ -1,7 +1,11 @@
...
-print("Mapping conforms with schema.")
+from .assessment_mapping import cof_r2w2_assessment as mapping
+from .assessment_mapping_schema import (
+    top_level_assessment_mapping_schema as schema,
+)
+
+
+def test_assessment_mapping_schema():
+    validate(instance=mapping, schema=schema)
+    print("Mapping conforms with schema.")
```
SentientHousePlant, line 11: ⛏️ Use a assert here. Validate raises an error if the the schema doesn't validate. Also pytest doesn't print strings by default.

SentientHousePlant, line 11:
```suggestion
def test_assessment_mapping_schema():
   try:
       validate(instance=mapping, schema=schema)
    except jsonschema.exceptions.ValidationError:
       assert False, "Schema not valid"
    else:
       assert True
```

SentientHousePlant: Github has messed up the tabbing but you get the gist.

SentientHousePlant (changes requested):

db/queries/assessment_records/queries.py
tferns: Thanks for the review!

This was a good shout, I've added it here: https://github.com/communitiesuk/funding-service-design-assessment-store/pull/45/commits/ff4aa11dbf05cb6f41bf0779f4efccc36d5a3c6d#diff-67a194b656540c4b2f66084a6833fa3d66fbc587d4824be2dc919d8286e5a536R119-R123

I don't think the `.one()` is required since we're using `.scalar(...)`

db/schemas/schemas.py
tferns: Have changed this here: https://github.com/communitiesuk/funding-service-design-assessment-store/commit/ff4aa11dbf05cb6f41bf0779f4efccc36d5a3c6d#diff-62e7f04d26309dfe4fe37ef06da26af5748ac7ed1346898667d2cffadc2b30d3

> This will mean that the we will need to add some more defers to the current queries.

Do you know where specifically? I don't see any others (at least on this branch) -- maybe something for after https://github.com/communitiesuk/funding-service-design-assessment-store/pull/46 is merged?

openapi/assessment_config.yml
tferns: Thanks! Added here: https://github.com/communitiesuk/funding-service-design-assessment-store/commit/ff4aa11dbf05cb6f41bf0779f4efccc36d5a3c6d#diff-ae0012ebd0e74874b167156eff37d9e47b8f7a7ec78ba7fb499078a7537c371d

db/queries/assessment_records/queries.py
tferns: Love a little type annotation: https://github.com/communitiesuk/funding-service-design-assessment-store/commit/ff4aa11dbf05cb6f41bf0779f4efccc36d5a3c6d#diff-67a194b656540c4b2f66084a6833fa3d66fbc587d4824be2dc919d8286e5a536R109

config/mappings/test_mapping_config.py
tferns: Yeah good shout, have also moved this to the `tests` dir

tferns:
Superseded by https://github.com/communitiesuk/funding-service-design-assessment-store/pull/50 (big refactor)

PR #46, opened 2022-11-24
fs-1580-score-app-sub-crit

gio-karunakaran:


db/queries/scores/queries.py
```diff
@@ -0,0 +1,63 @@
...
+) -> Dict:
+    """get_just_score_for_application_sub_crit executes a query on scores
+    which returns the most recent score for the given application_id and 
+    sub_criteria_id.
+
+    :param application_id: The stringified application UUID.
+    :param sub_criteria_id: The stringified sub_criteria UUID.
+    :return: dictionary.
+    """
+    stmt = select(JustScore).where(
```
ramsharma-prog, line 29: can stmt be more descriptive?

db/queries/scores/queries.py
```diff
@@ -0,0 +1,63 @@
...
+from db import db
+from db.models.score import JustScore
+from db.queries.assessment_records._helpers import derive_values_from_json
+from db.schemas import JustScoreMetadata
+from sqlalchemy import func
+from sqlalchemy import select
+from sqlalchemy.orm import defer
+
+
+def get_just_score_for_application_sub_crit(
```
SentientHousePlant, line 18: ⛏️ This needs to be more accurate - it selects the most recent score.

db/queries/scores/queries.py
```diff
@@ -0,0 +1,63 @@
...
+    sub_criteria_id.
+
+    :param application_id: The stringified application UUID.
+    :param sub_criteria_id: The stringified sub_criteria UUID.
+    :return: dictionary.
+    """
+    stmt = select(JustScore).where(
+        JustScore.application_id == application_id,
+        JustScore.sub_criteria_id == sub_criteria_id
+        ).order_by(JustScore.timestamp.desc())
```
SentientHousePlant, line 32: 💯 Good to see the modern SQLAlchemy syntax being used. This will ensure forward compatibility and is much easier to understand then the old syntax. It is now the recommended syntax (see https://docs.sqlalchemy.org/en/14/tutorial/ for the details on the 1.4/2 syntax in SQLA)

db/queries/scores/queries.py
```diff
@@ -0,0 +1,63 @@
...
+
+    latest_score_metadata = db.session.scalar(stmt)
+    metadata_serialiser = JustScoreMetadata()
+    latest_score_metadata = metadata_serialiser.dump(latest_score_metadata)
+
+    return latest_score_metadata
+
+
+def post_just_score_for_application_sub_crit(
```
SentientHousePlant, line 41: ⛏️ This function doesn't post - it adds a row to the db :)

db/queries/scores/queries.py
```diff
@@ -0,0 +1,63 @@
...
+    sub_criteria_id: str, user_id: str
+) -> Dict:
+    """get_just_score_for_application_sub_crit executes a query on scores
+    which returns the most recent score for the given application_id and 
+    sub_criteria_id.
+
+    :param application_id: The stringified application UUID.
+    :param sub_criteria_id: The stringified sub_criteria UUID.
+    :return: dictionary.
+    """
```
SentientHousePlant, line 52: 🤩 Gorgeous docstrings! Shame this one is copy and pasted from the one above ;)

db/schemas/schemas.py
```diff
@@ -15,3 +16,11 @@ class Meta:
 
     workflow_status = Enum(Status)
     language = Enum(Language)
+
+
+class JustScoreMetadata(SQLAlchemyAutoSchema):
```
SentientHousePlant, line 21: 🤔 If you like, you can specify in what format to return the DateTimes - see https://marshmallow.readthedocs.io/en/stable/marshmallow.schema.html#marshmallow.schema.Schema.Meta and in particular the `timedateformat` option.

api/routes/score_routes.py
```diff
@@ -0,0 +1,48 @@
...
+        application_id=application_id, sub_criteria_id=sub_criteria_id
+    )
+
+    return latest_score
+
+
+def post_score_for_application_sub_criteria(
+    score: int, justification: str, application_id: str, timestamp: str, 
+    sub_criteria_id: str, user_id: str
+) -> Dict:
```
SentientHousePlant, line 32: 💯 I like that the kwargs are listed here - I think just putting **kwargs here hides details.

db/models/assessment_record/assessment_records.py
```diff
@@ -57,6 +58,8 @@ class AssessmentRecord(db.Model):
         Computed(func.md5(cast(jsonb_blob, TEXT)), persisted=True),
     )
 
+    scores = relationship("JustScore")
```
SentientHousePlant, line 61: ⛏️ Make sure that this change is reflected in the queries for `assessment_records`, I can assist with this :)

tasks/db_tasks.py
```diff
@@ -55,7 +55,7 @@ def seed_dev_db(c):
 
         with app.app_context():
 
-            from tests.conftest import seed_database
+            from tests.conftest import seed_database_randomly
```
SentientHousePlant, line 58: 💯  Thank you for fixing this.

SentientHousePlant (changes requested):
Great stuff Gio! :) Some comments to keep you on your toes.

SentientHousePlant:
Also...tests!!

db/models/score/scores.py
```diff
@@ -0,0 +1,40 @@
...
+    application_id = db.Column(
+        "application_id", UUID, ForeignKey("assessment_records.application_id")
+    )
+
+    timestamp = db.Column(
+        "timestamp", db.DateTime(), nullable=False
+    )
+
+    sub_criteria_id = db.Column(
+        "sub_criteria_id", db.String(),  nullable=False
```
gidsg, line 37: According to the modal on Mural this should be a UUID rather than a String. I'm not sure where the details about the subcriteria are held though?

db/models/score/scores.py
```diff
@@ -0,0 +1,40 @@
...
+from db import db
+from sqlalchemy import ForeignKey 
+from sqlalchemy.dialects.postgresql import UUID
+
+
+class JustScore(db.Model):
+    """Score The sqlalchemy-flask model class used to define the
+    `scores` table in the Postgres database."""
+
+    __tablename__ = "just_scores"
```
gidsg, line 14: Can we not use abbreviations in table names? So I would prefer `scores` in this case and the class called `Score`.

api/routes/score_routes.py
```diff
@@ -0,0 +1,48 @@
...
+
+    latest_score = get_just_score_for_application_sub_crit(
+        application_id=application_id, sub_criteria_id=sub_criteria_id
+    )
+
+    return latest_score
+
+
+def post_score_for_application_sub_criteria(
+    score: int, justification: str, application_id: str, timestamp: str,
```
gidsg, line 30: Can `timestamp` be a DateTime rather than a string?

db/models/score/scores.py
```diff
@@ -0,0 +1,40 @@
...
+
+    justification = db.Column(
+        "justification", db.Text(), nullable=False
+    )
+
+    application_id = db.Column(
+        "application_id", UUID, ForeignKey("assessment_records.application_id")
+    )
+
+    timestamp = db.Column(
```
gidsg, line 32: Can we make this `date_created` rather than `timestamp`?

db/models/score/scores.py
gidsg: CC @Adam-W1 does it make more sense for this to be a string, given the config dict?

db/queries/scores/queries.py
gio-karunakaran: stmt is just the sql query that is used to parse into the search and seems to be best practice to use as in sqlalchemy: https://docs.sqlalchemy.org/en/14/orm/queryguide.html

db/queries/scores/queries.py
gio-karunakaran: done!

gidsg, line 18: Can we call this `get_current_score` if that's what it's doing?

db/migrations/versions/91e0d528fc9f_.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('just_scores',
+    sa.Column('score_id', postgresql.UUID(as_uuid=True), nullable=False),
+    sa.Column('score', sa.Integer(), nullable=False),
+    sa.Column('justification', sa.Text(), nullable=False),
+    sa.Column('application_id', postgresql.UUID(), nullable=True),
+    sa.Column('timestamp', sa.DateTime(), nullable=False),
```
gidsg, line 27: Can we call this `date_created`

gidsg: Think it is safe to default this as per https://github.com/communitiesuk/funding-service-design-application-store/blob/559d62e7f08f1dc5b143283f2110fc46014803d3/db/models/applications.py#L41 and then there is no need for the client to provide it.

db/queries/scores/queries.py
gio-karunakaran: haha have sorted this now :) 👍🏾

api/routes/score_routes.py
gio-karunakaran: I'm not sure if I want to hide the details? I was thinking it would be nice to see what this func takes when investigating or for debugging etc?

db/models/assessment_record/assessment_records.py
gio-karunakaran: Which queries would need updating in assessment records?

db/models/score/scores.py
gio-karunakaran: @Adam-W1 @gidsg yes! i was also confused by this, looking at the config dict i thought the most sensible thing to do was to keep it as a string otherwise sqlalchemy would complain about it not being a properly formed UUID 🤔

db/models/score/scores.py
gio-karunakaran: ahhh, so i called this justscores as it is what was being used in assessment frontend, can change this if score is still preferred?

db/queries/scores/queries.py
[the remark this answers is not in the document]
gio-karunakaran: because I am using method scalar() this will always return one result :) and in terms of sorting, I'm not 100% sure if we're always going to be sorting by either DESC or ASC but looks very cool!! :D - something to consider when we know more about score histories etc... ;)

db/models/score/scores.py
gio-karunakaran: have changed this now 👍🏾

db/queries/scores/queries.py
gio-karunakaran: have changed this now 👍🏾

db/migrations/versions/91e0d528fc9f_.py
gio-karunakaran: ahh yes great shout! have added this now 👍🏾

db/migrations/versions/f783496def21_.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('just_scores',
+    sa.Column('score_id', postgresql.UUID(as_uuid=True), nullable=False),
+    sa.Column('score', sa.Integer(), nullable=False),
+    sa.Column('justification', sa.Text(), nullable=False),
+    sa.Column('application_id', postgresql.UUID(), nullable=True),
+    sa.Column('date_created', sa.DateTime(), nullable=True),
```
SentientHousePlant, line 26: 🤔 Should this have a default? like func.now()?

db/models/score/scores.py
```diff
@@ -0,0 +1,41 @@
...
+within the Postgres db.
+"""
+import uuid
+from db import db
+from sqlalchemy import ForeignKey
+from sqlalchemy.sql import func
+from sqlalchemy.dialects.postgresql import UUID
+
+
+class JustScore(db.Model):
```
SentientHousePlant, line 11: ⛏️ Make model and table name clearer - ie don't abbreviate.

db/queries/scores/__init__.py
```diff
@@ -0,0 +1,8 @@
+# flake8: noqa
+from .queries import get_latest_just_score_for_application_sub_crit
+from .queries import create_just_score_for_application_sub_crit
+
+__all__ = [
+    get_latest_just_score_for_application_sub_crit,
+    create_just_score_for_application_sub_crit,
```
SentientHousePlant, line 7: ⛏️ `__all__` takes a list of strings.
```suggestion
    "get_latest_just_score_for_application_sub_crit",
    "create_just_score_for_application_sub_crit",
```

db/queries/scores/queries.py
```diff
@@ -0,0 +1,67 @@
...
+    stmt = select(JustScore).where(
+        JustScore.application_id == application_id,
+        JustScore.sub_criteria_id == sub_criteria_id
+        ).order_by(JustScore.date_created.desc())
```
SentientHousePlant, line 32: ⛏️ Use a limit here so we don't select every score.

db/queries/scores/queries.py
```diff
@@ -0,0 +1,67 @@
...
+
+    latest_score_metadata = db.session.scalar(stmt)
+    metadata_serialiser = JustScoreMetadata()
+    latest_score_metadata = metadata_serialiser.dump(latest_score_metadata)
```
SentientHousePlant, line 36: ⛏️ Instead of reusing a variable name consider having a `row` and a `json` variable.

db/queries/scores/queries.py
```diff
@@ -0,0 +1,67 @@
...
+    :param justification: The justification text.
+    :param date_created: The date_created.
+    :param user_id: The stringified user_id.
+    :return: dictionary.
+    """
+    score = JustScore(
+        score=score, justification=justification, application_id=application_id, 
+        sub_criteria_id=sub_criteria_id,user_id=user_id
+    )
+    db.session.add(score)
```
SentientHousePlant, line 61: ⛏️ Consider using a [insert](https://docs.sqlalchemy.org/en/14/tutorial/data_insert.html).

db/schemas/schemas.py
```diff
@@ -15,3 +17,16 @@ class Meta:
...
+    """JustScoreMetadata The marshmallow class used to turn SQLAlchemy
+    rows into json for return in http responses."""
+
+    class Meta:
+        model = JustScore
+        include_relationships = True
```
SentientHousePlant, line 28: ❓ What relationship are we including here?

db/schemas/schemas.py
```diff
@@ -15,3 +17,16 @@ class Meta:
...
+        load_instance = True
+    
+    application_id = auto_field(dump_only=True)
```
SentientHousePlant, line 31: ⛏️ You can include foreign keys by setting a [parameter in the Meta. ](https://marshmallow-sqlalchemy.readthedocs.io/en/latest/api_reference.html#marshmallow_sqlalchemy.SQLAlchemyAutoSchemaOpts)

SentientHousePlant (changes requested):

db/queries/scores/queries.py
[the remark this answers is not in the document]
SentientHousePlant: Yes, but SQLA will just take a single value from whatever is returned by postgres - there will still be quite a heavy load on the database. Do check the sql output in pytest and see what statement is emitted here.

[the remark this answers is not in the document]
SentientHousePlant: but yes scalar is better then one, good spot :)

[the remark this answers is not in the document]
SentientHousePlant: and the index can be used regardless of ASC or DESC, but you if you know then you can specify to make the index super small, if not it doesn't matter....**tl;dr** add the index

db/migrations/versions/f783496def21_.py
gio-karunakaran: i thought it was weird it didn't show here... it is func.now() in the model 👍🏾 not sure why there's nothing that shows this in the migration, have tested this and date_created does get auto generated 👌🏾

SentientHousePlant: have you got it set as the `server_default`?

db/models/score/scores.py
SentientHousePlant: If you use `UUID(as_uuid=True)` it might not complain anymore.

db/migrations/versions/f783496def21_.py
gidsg: Ah, I see @SentientHousePlant's point, we can set a default in the migration itself e.g. as per https://github.com/communitiesuk/funding-service-design-application-store/blob/559d62e7f08f1dc5b143283f2110fc46014803d3/db/models/applications.py#L41

gio-karunakaran: ahh thank you @SentientHousePlant @gidsg that makes sense, all done now 👌🏾

tests/test_db_function.py
```diff
@@ -53,3 +58,43 @@ def test_non_blob_columns_mutable(db_session):
...
+        "sub_criteria_id": sub_criteria_id,
+        "score": 3,
+        "justification": "bang average",
+        "user_id": "test"
+    }
+    score_metadata = create_just_score_for_application_sub_crit(**assessment_payload)
+
+    assert len(score_metadata) == 7
+    assert score_metadata.get("date_created")
+    assert score_metadata.get("score") == 3
```
SentientHousePlant, line 79: ⛏️ Don't use `get` when an object must have this property. It allows for `None`s to propagate though the code.

tests/test_db_function.py
```diff
@@ -53,3 +58,43 @@ def test_non_blob_columns_mutable(db_session):
         db_session.commit()
     except sqlalchemy.exc.InternalError:
         raise AssertionError
+
+
+def test_create_scores_for_application_su_crit():
```
SentientHousePlant, line 63: 📜 Docstring

tests/test_db_function.py
```diff
@@ -53,3 +58,43 @@ def test_non_blob_columns_mutable(db_session):
...
+        "user_id": "test"
+    }
+    score_metadata = create_just_score_for_application_sub_crit(**assessment_payload)
+
+    assert len(score_metadata) == 7
+    assert score_metadata.get("date_created")
+    assert score_metadata.get("score") == 3
+
+
+def test_scores_for_application_sub_crit():
```
SentientHousePlant, line 82: 📜 Docstring

SentientHousePlant (changes requested):

db/models/score/scores.py
gio-karunakaran: have changed from JustScore to Score 👍🏾

db/queries/scores/queries.py
gio-karunakaran: I think we mentioned this before? scalar() will always pick a maximum of 1 element already so the limit is already applied?

db/queries/scores/queries.py
gio-karunakaran: good shout!

db/queries/scores/queries.py
gio-karunakaran: is the difference between add and insert that with insert you don't need to commit after? 🤔

db/schemas/schemas.py
gio-karunakaran: I think by default it includes all relationships within the model which for this case is the foreign key relationship with assessment records 👍🏾

db/queries/scores/queries.py
SentientHousePlant: I don't think the sql query will not contain this limit.

SentientHousePlant: check the emitted sql query and see if a limit is there - if not it will be worth adding one.

db/queries/scores/queries.py
SentientHousePlant: You still need to commit but it's more modern sqlalchemy.

tests/test_db_function.py
gio-karunakaran: another great shout!

db/queries/scores/queries.py
gio-karunakaran: no limit emitted from the query so have added that in there now 👍🏾

db/models/__init__.py
```diff
@@ -1,4 +1,5 @@
 # flake8: noqa
 from db.models.assessment_record import AssessmentRecord
+from db.models.score import Score
 
-__all__ = [AssessmentRecord]
+__all__ = [AssessmentRecord, Score]
```
SentientHousePlant, line 5: these should be strings

db/models/score/__init__.py
```diff
@@ -0,0 +1,2 @@
+# flake8: noqa
+from .scores import Score
```
SentientHousePlant, line 2: Include a `__all__`

db/models/score/scores.py
SentientHousePlant: @gio-karunakaran did you manage to fix this?

db/models/score/scores.py
```diff
@@ -0,0 +1,41 @@
...
+
+class Score(db.Model):
+    """Score The sqlalchemy-flask model class used to define the
+    `scores` table in the Postgres database."""
+
+    __tablename__ = "scores"
+
+    id = db.Column(
+        "score_id", UUID(as_uuid=True), default=uuid.uuid4, primary_key=True
+    )
```
SentientHousePlant, line 19: Using `sqlalchemy`s Column has better autocomplete and doesn't change the migrations.

tests/test_db_function.py
```diff
@@ -53,3 +58,49 @@ def test_non_blob_columns_mutable(db_session):
...
+        "application_id": application_id,
+        "sub_criteria_id": sub_criteria_id,
+        "score": 3,
+        "justification": "bang average",
+        "user_id": "test"
+    }
+    score_metadata = create_score_for_application_sub_crit(**assessment_payload)
+
+    assert len(score_metadata) == 7
+    assert score_metadata["date_created"]
```
SentientHousePlant, line 81: Might be better to use ` "data_created" in score_metdata ` so its more clear what the check is.

db/schemas/__init__.py
```diff
@@ -1,3 +1,4 @@
 from .schemas import AssessmentRecordMetadata
+from .schemas import ScoreMetadata
 
-__all__ = [AssessmentRecordMetadata]
+__all__ = [AssessmentRecordMetadata, ScoreMetadata]
```
SentientHousePlant, line 4: strings

SentientHousePlant (approved):
Some small changes are requested, but I won't need to review them since they are stylistic rather than functional. 

Thanks for responding to my (many) picky changes and writing clean looking code! Thank you! <3

PR #50, opened 2022-11-28
FS-1581: Assessor Task List page state endpoint

tferns:
**Summary:** 

- Add endpoint to retrieve state for populating the front-end assessor task list page
- Consumed by: https://github.com/communitiesuk/funding-service-design-assessment/pull/58

**Note:** Supersedes https://github.com/communitiesuk/funding-service-design-assessment-store/pull/45


tasks/db_tasks.py
```diff
@@ -98,7 +98,7 @@ def seed_dev_db(c):
                 f" {apps * rounds * funds} test rows."
             )
 
-            seed_database(apps, rounds, funds)
+            seed_database_randomly(apps, rounds, funds)
```
gidsg, line 101: Why are we making this change in this PR? Seems unrelated?

db/schemas/schemas.py
```diff
@@ -11,7 +12,47 @@ class AssessmentRecordMetadata(SQLAlchemyAutoSchema):
...
+        )
+
+    @post_dump
+    def discard_irrelevant_fields(self, item, many, **kwargs):
+        """Discard irrelevant fields from the jsonb_blob."""
+        item["date_submitted"] = item["jsonb_blob"]["date_submitted"]
+        item["project_reference"] = item["short_id"]
+        del item["short_id"]
+        del item["jsonb_blob"]
+        return item
```
SentientHousePlant, line 58: Why are we doing this?

db/schemas/schemas.py
```diff
@@ -11,7 +12,47 @@ class AssessmentRecordMetadata(SQLAlchemyAutoSchema):
...
+    - round_id
+    """
+
+    class Meta:
+        model = AssessmentRecord
+        exclude = tuple(
+            field
+            for field in AssessmentRecord.__table__.columns.keys()
+            if field not in ASSESSOR_TASK_LIST_METADATA_FIELDS
+        )
```
SentientHousePlant, line 49: You can use `include` when you create the serialiser object - IE `serialiser = AssessmentRecordSchema(include=("short_id", "data_submitted"))`

SentientHousePlant: This way you don't have to write new objects for exclusion

db/schemas/schemas.py
SentientHousePlant: NB: see key and field here https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.Field . This might be a cleaner way of doing this. Also, we extract information from the json blob on insert so we don't have to do it here

db/schemas/schemas.py
SentientHousePlant: If we are missing a column add it and update the mapper here https://github.com/communitiesuk/funding-service-design-assessment-store/blob/main/db/queries/assessment_records/_helpers.py#L7

db/schemas/__init__.py
```diff
@@ -1,3 +1,4 @@
 from .schemas import AssessmentRecordMetadata
+from .schemas import AssessorTaskListMetadata
 
-__all__ = [AssessmentRecordMetadata]
+__all__ = ["AssessmentRecordMetadata", "AssessorTaskListMetadata"]
```
SentientHousePlant, line 4: 💯

api/routes/assessment_routes.py
```diff
@@ -22,3 +25,28 @@ def all_assessments_for_fund_round_id(
...
+    :return: A dictionary.
+    """
+
+    metadata = find_assessor_task_list_state(application_id)
+    sections, criterias = transform_to_assessor_task_list_metadata(
+        metadata["fund_id"], metadata["round_id"]
+    )
+
+    # We don't need to return this state.  It's only used to get the config above.
+    del metadata["round_id"]
```
SentientHousePlant, line 44: `del` is not typically used in python for filter since it's not very explicit and can create issues because dictionarys are typically passed by reference rather then value - i'd recommand writing a dictionary comprehension which filters and creates a new filtered dict to return.

SentientHousePlant (changes requested):

tasks/db_tasks.py
tferns: Think it was re-named but not committed in another PR, I added it so I could run the script and committed it at the same time for anyone else running it (incase they hit the same issue)

db/schemas/schemas.py
tferns: I was using that originally but then at some point decided to go with this, I think it's when I (at least at the time) thought I had to write a new class and couldn't achieve what I was trying purely by passing properties (the `@post_dump` thing) - I'll take a look at re-using `include` if I can remove the `@post_dump` utilizing the docs you mentioned above

db/schemas/schemas.py
tferns: I'll take a deeper look at this tomorrow

tferns: Have tried to utilize fields now and removed the ugly `post_dump` stuff.  If you know a better way to do this please let me know, admittedly not too familiar with this library.

api/routes/assessment_routes.py
tferns: Thanks for noticing that, have refactored in https://github.com/communitiesuk/funding-service-design-assessment-store/pull/50/commits/cf75286e37b0cfe1f95423e8aea0ae793ec6502d

db/schemas/schemas.py
tferns: Have changed to use `only` as an argument in the constructor, again if you know a better way please let me know, thanks for the feedback 😄

db/schemas/__init__.py
Adam-W1: just out of interest folks, is good practice pythonista stuff? My understanding was that the default 'import *' behaviour is to import all symbols that do not begin with an underscore anyway?

api/routes/assessment_routes.py
```diff
@@ -22,3 +25,26 @@ def all_assessments_for_fund_round_id(
...
+def get_assessor_task_list_state(application_id: str) -> dict:
+    """get_assessor_task_list_state Function used by the endpoint
+    `/assessor_task_list/{application_id}`.
+
+    :param application_id: The stringified application UUID.
+    :return: A dictionary.
+    """
+
+    metadata = find_assessor_task_list_state(application_id)
+    sections, criterias = transform_to_assessor_task_list_metadata(
```
Adam-W1, line 39: 👌

config/mappings/assessment_mapping_fund_round.py
```diff
@@ -0,0 +1,47 @@
...
+                    "theme_count": len(sc["themes"]),
+                    "status": "Not started",  # TODO: Get dynamic state.
+                }
+                for sc in c["sub_criteria"]
+            ],
+        }
+        for c in mapping["scored_criteria"]
+    ]
+
+    return sections, criterias
```
Adam-W1, line 47: Would be good to see some testing for this

db/queries/assessment_records/queries.py
```diff
@@ -105,3 +108,44 @@ def find_answer_by_key_runner(field_key: str, app_id: str) -> List[tuple]:
         .filter(AssessmentRecord.application_id == app_id)
         .one()
     )
+
+
+def find_assessor_task_list_state(application_id: str) -> dict:
```
Adam-W1, line 113: Might we want to return tasklist specific state here. For example, we only want criteria scoring and statuses, so not sure if we need the json blob of answers.

tferns: I think so too yeah.  Will need to find out when that work can actually be done, as I think there are in-flight stories at the moment making changes to the data model?  I'll actually ask Gio to get confirmation on this, because if not I might be able to do TDD on the current data model

config/mappings/assessment_mapping_fund_round.py
tferns: I've added https://github.com/communitiesuk/funding-service-design-assessment-store/pull/50/commits/f03237ce25be02fc3eadc29f7c9699c84b9310fa which sort of acts as a test/documentation for the eventually transformed structure (this is what is returned from the endpoint, the dataclass `AssessorTaskList` is built form, and used to dynamically populate the UI with config elements, and state (once part of the transformation)

db/queries/assessment_records/queries.py
```diff
@@ -105,3 +108,44 @@ def find_answer_by_key_runner(field_key: str, app_id: str) -> List[tuple]:
...
+    assessment_record = db.session.scalar(stmt)
+
+    assessment_record_json = AssessorTaskListMetadata(
+        only=(
+            "short_id",
+            "project_name",
+            "date_submitted",
+            "workflow_status",
+            "fund_id",
+            "round_id",
```
SentientHousePlant, line 147: should the blob be excluded here? If so, then we shouldn't load it on line 127.

db/queries/assessment_records/queries.py
tferns: Have created https://digital.dclg.gov.uk/jira/browse/FS-2000 for when the `score` data model changes are in place and that work can be picked up 👍

tferns: Have created https://digital.dclg.gov.uk/jira/browse/FS-2000 for when the score data model changes are in place and that work can be picked up 👍

SentientHousePlant (approved):

db/schemas/__init__.py
SentientHousePlant: __all__ specifies what happens when you use *. While we don't use * often, many libraries do (like pdoc) and it keeps the public facing parts very clear. It's also used by pdoc to figure out what are the python things worth documenting.

SentientHousePlant (changes requested):
😢 sorry to rollback my approval - but if we are adding a new route then we should test it.

- [x] Add test for new endpoint.

tferns:
> 😢 sorry to rollback my approval - but if we are adding a new route then we should test it.

No worries makes sense!  I'll add some test coverage for this so we're more confident it works as intended.

tferns:
@SentientHousePlant 
> 😢 sorry to rollback my approval - but if we are adding a new route then we should test it.

I've pushed https://github.com/communitiesuk/funding-service-design-assessment-store/pull/50/commits/dfc0c5e56f6b7f237041367947a9bcf97f3e0c47 with a test for the route and the database query, let me know if you have any other feedback  👍

config/mappings/assessment_mapping_fund_round.py
```diff
@@ -0,0 +1,47 @@
...
+    cof_r2w2_assessment_mapping as cof_r2w2,
+)
+from fsd_utils import CommonConfig
+
+fund_round_to_assessment_mapping = {
+    f"{CommonConfig.COF_FUND_ID}:{CommonConfig.COF_ROUND_2_ID}": cof_r2w2,
+}
+
+
+def transform_to_assessor_task_list_metadata(
```
Adam-W1, line 11: Eventually this function will be making calls to the databse as well as config, so I wonder if this might be better placed in the route controller for returning assessment metadata

Adam-W1 (commented):
look great! just a couple small comments to respond to 👍

tferns:
> looks great! just a couple small comments to respond to 👍

Hey, thanks again for the review.  I can only see one new comment, unless you're referring to any old comments? 

I've added https://github.com/communitiesuk/funding-service-design-assessment-store/pull/50/commits/ddf7dfeab8c4f7e67726db1d56d661c76fa8f309 to better place the transformation logic.  Let me know if you have any other suggestions! Thanks

SentientHousePlant (approved):
Beautiful stuff Thomas! Thanks for taking the time to respond to my comments.

PR #52, opened 2022-12-01
bau-linting-checks

SentientHousePlant:
- Introduces the pre-commit check to github actions. Preventing people ignoring pre-commit checks by....not using it.
- Formats existing codebase using pre-commit.


tferns (approved):
Nice! I like the concept of this since I'm a fan of pre-commit, however the way the README.md reads is that it's more of an optional tool:

 > This repo comes with a .pre-commit-config.yaml, if you wish to use this do the following while in your virtual enviroment:
 
Could we maybe update that just to give people a better indication their code should follow the more opinionated standards enforced by our pre-commit hooks?  Not that it really matters, since they will find out anyways with a failing build email 😄

Also, pre-commit comes in simple to use CI form (specifically for GitHub) which works based upon our existing config, see [pre-commit.ci](https://pre-commit.com/) or [pre-commit.ci/lite](https://pre-commit.ci/lite.html) - it might be good to add that instead so it fixes things for people, instead of a more manual `push > fail > fix > push` cycle?

Well, I'm sure after a few failures people will cave and install the hooks anyways 😅

Approving cause I like the idea, but food for thought above ^^^

SentientHousePlant:
@tferns Great suggestion! @gidsg is it possible to install the `lite` version of pre-commit? :) Then we can just add that line to our action.

PR #54, opened 2022-12-02
Fs 1929 nav sidebar

Adam-W1:
https://digital.dclg.gov.uk/jira/browse/FS-1929

The navigation sidebar when viewing an assessment's sub criteria.

Linked to: https://github.com/communitiesuk/funding-service-design-assessment/pull/61

TODO: when a scoring table has been implemented, the 'score_submitted' boolean can be queried and returned on the sub_criterias return payload.


tests/test_routes.py
```diff
@@ -71,3 +72,30 @@ def test_search(client):
...
+def test_get_sub_criteria(request, client, sub_criteria_response_keys):
+    """Test to check that sub criteria metadata and ordered themes are returned for
+    a COFR2W2 sub criteria"""
+
+    sub_criteria_id = "benefits"
+    response_json = client.get(
+        f"/sub_criteria_overview/{sub_criteria_id}"
+    ).json
+
+    assert sub_criteria_response_keys in response_json
```
gidsg, line 87: If we care about the order here, should we assert on the keys directly as an array rather than doing a substring match in the JSON?

Adam-W1: Yes you're right! I've added another check to ensure we are ordering the themes as expected.

gidsg (approved):
LGTM 👍

api/models/sub_criteria.py
```diff
@@ -0,0 +1,32 @@
+import inspect
+from dataclasses import dataclass
+from typing import List, Dict
+from api.models.theme import Theme
+from stringcase import lowercase
+from dataclasses import dataclass
+from dataclass_dict_convert import dataclass_dict_convert
+
+
+@dataclass_dict_convert()
```
tferns, line 10: This annotation is pretty cool - looking at [the docs](https://pypi.org/project/dataclass-dict-convert/) it looks like it does it for nested lists, are the `__post_init__` and `from_filtered_dict` still necessary?

api/models/theme.py
```diff
@@ -0,0 +1,30 @@
...
+    @classmethod
+    def from_filtered_dict(cls, d: dict):
+        # Filter unknown fields from JSON dictionary
+        return cls(
+            **{
+                k: v
+                for k, v in d.items()
+                if k in inspect.signature(cls).parameters
+            }
+        )
```
tferns, line 30: Same question as above

api/routes/subcriterias/get_sub_criteria.py
```diff
@@ -0,0 +1,45 @@
...
+            lambda sub_criteria: sub_criteria["id"] == sub_criteria_id,
+            sub_criterias,
+        )
+    )
+    if len(matching_sub_criteria) == 1:
+        return matching_sub_criteria[0]
+    elif len(matching_sub_criteria) > 1:
+        msg="sub_criteria: '{sub_criteria_id}' duplicated."
+        current_app.logger.error(msg)
+        raise ValueError(msg)
```
tferns, line 31: Does this ever happen or just precautionary?

tests/test_routes.py
```diff
@@ -71,3 +72,35 @@ def test_search(client):
         ).json
 
         assert len(response_json) == len(rows)
+
+
+@pytest.mark.parametrize("sub_criteria_response_keys", ["id", "name", "score_submitted","themes"])
```
tferns, line 77: Bit of a nitpick, but this parametrize seems unnecessary since we just have a single input

tferns (approved):
Looks good.  Question as to whether some dataclass methods are required

api/routes/subcriterias/get_sub_criteria.py
Adam-W1: thats correct, it should never happen but this gives us a better trace if anything slips through in future

api/models/sub_criteria.py
Adam-W1: So the post_init and from_filtered_dict work together to ensure we can create class instances down through every linked subclass (from a dictionary), without the post_init we dont get the nested lists as a dataclass (and this then nobbles the dict_convert :[).

tests/test_routes.py
Adam-W1: I think my plural naming misleads here, but each time we are only testing for a single response key, so we need to parametrize to test for each key. Good spot ill change the name to reflect this.

PR #56, opened 2022-12-07
BAU - Adding total funding requested to assessment overview and moving business case to unscored

stevewalton28:
1. Adding total funding requested to assessment overview return to support new assessor dashboard design
2. Moving business case from Management Case for COF config to Unscored in the unscored assessment structure to support COF business decision to make business case for information only and have assessment on structured data collected via forms

### DIFF
--- a/config/envs/default.py
+++ b/config/envs/default.py
@@ -38,6 +38,8 @@ class DefaultConfig:
         "postgres://", "postgresql://"
     )
 
+    # SQLALCHEMY_DATABASE_URI = "postgresql://postgres:password@localhost:5432/assessment_store"
+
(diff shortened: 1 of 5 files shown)


tferns (approved):
Looks good but there are failing tests.

Two of them aren't really relevant/useful, if the config is going to be changing:

`test_transform_to_assessor_task_list_metadata`
`test_get_application_metadata_for_application_id`

The one I think is worth keeping:

 `test_find_assessor_task_list_ui_metadata` 
 
 Its an integration test with the database and checks that the query works, it's failing cause it's doing an equals comparison to a dict and you've added a new field

PR #57, opened 2022-12-08
FS-2045: fix manifest.yml

gidsg:
theory is that having 2 applications keys was causing the manifest to be interpreted in a weird way and databases being bound incorrectly


harryyo (approved):
👍

PR #72, opened 2022-12-21
FS-2102-update-comments-endpoint

gio-karunakaran:


adamdavies1 (approved):
LGTM

tferns (approved):
looks good, only concern is if the tests are running on a test db then over time it'll get filled up... might not be something we need to worry about though just a thought

PR #75, opened 2022-12-22
FS-2014: Add endpoint to retrieve banner state

tferns:
[Add endpoint to retrieve banner state](https://github.com/communitiesuk/funding-service-design-assessment-store/pull/75/commits/8279607efe5281a364be1466a47740d429ba48d4)
- Update metadata to use Enum as was seeing error

[Add flag routes and database model/migration](https://github.com/communitiesuk/funding-service-design-assessment-store/pull/75/commits/2f1a616e6145b90c15dc4ad863ebfa58d967f992)
- Get request to return all flags for a given application
- Post request to create a flag for a given application
- Migration script introduced to create the flags table in db

### DIFF
--- a/db/models/flags/flags.py
+++ b/db/models/flags/flags.py
@@ -1,0 +1,43 @@
+"""The module containing all code related to the `flags` table
+within the Postgres db.
+"""
+import uuid
+
+from db import db
+from db.models.flags.enums import FlagType
+from db.models.flags.enums import ResolutionType
+from sqlalchemy import ForeignKey
+from sqlalchemy.dialects.postgresql import ENUM
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.sql import func
+
+
+class Flag(db.Model):
+    """Flag: The sqlalchemy-flask model class used to define the
+    `flags` table in the Postgres database."""
+
+    __tablename__ = "flags"
+
+    id = db.Column(
+        "flag_id", UUID(as_uuid=True), default=uuid.uuid4, primary_key=True
+    )
+
+    justification = db.Column("justification", db.Text(), nullable=False)
+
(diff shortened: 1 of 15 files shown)


db/models/flags/flags.py
```diff
@@ -0,0 +1,43 @@
...
+
+    section_to_flag = db.Column("section_to_flag", db.Text(), nullable=False)
+
+    flag_type = db.Column("flag_type", ENUM(FlagType), nullable=False)
+
+    resolution_reason = db.Column(
+        "resolution_reason", ENUM(ResolutionType), nullable=True
+    )
+
+    application_id = db.Column(
```
Adam-W1, line 35: just wondering if there is a restraint that there should only ever be one flag per application at a time? Unsure on the business rules around this so more of a Q tbf

tferns: Hey, good point!  From what I could gather, there _might_ be different types of flagging in the future... so maybe something for the future?

openapi/api.yml
```diff
@@ -155,6 +157,67 @@ paths:
...
+      responses:
+        201:
+          description: SUCCESS - Flag for Application Submitted
+          content:
+            application/json:
+              schema:
+                type: object
+                items:
+                  $ref: 'components.yml#/components/schemas/Flag'
+        401:
```
adamdavies1, line 210: Is this Supposed to ve "Could not Save Flag"?

adamdavies1 (approved):
LGTM

openapi/api.yml
tferns: Good eye, have fixed in https://github.com/communitiesuk/funding-service-design-assessment-store/pull/75/commits/b6b34c0ed88dcf65c191572addf4a181fa67a8c3

PR #82, opened 2023-01-06
FS-2111: Create migration to introduce new locationJsonBlob field

johnoneill46:
https://digital.dclg.gov.uk/jira/browse/FS-2111

As part of FS-2111 we require location data to be stored into the assessment_store DB as a new blob field

Changes to populate this field will be done as part of this PR: https://github.com/communitiesuk/funding-service-design-assessment-store/pull/74


srh-sloan (approved):
LGTM 👍

PR #86, opened 2023-01-06
move to python script

Adam-W1:
https://digital.dclg.gov.uk/jira/browse/FS-2135

The previous implementation of the import script is not currently working with the application store data. To provide a higher level of developer confidence and control, this script is to be re-written (from SQL/Flask SQL-alchemy) based parsing of form JSON to python based parsing.


db/queries/assessment_records/_helpers.py
```diff
@@ -22,58 +25,70 @@
...
+#         for column_name, json_path in mapper.items()
+#     ]
+#     extract_fields_stmt = select(*selects).select_from(loaded_blob)
+#     query_result = db.session.execute(extract_fields_stmt).one()._asdict()
+#     return query_result
+
+def map_json(json):
+    asset_type = jsonpath_rw_ext.parse('$.forms[*].questions[*].fields[?(@.key == "yaQoxU")]').find(json)[0].value["answer"]
+    funding_one = jsonpath_rw_ext.parse('$.forms[*].questions[*].fields[?(@.key =="JzWvhj")]').find(json)[0].value["answer"]
+    funding_two = jsonpath_rw_ext.parse('$.forms[*].questions[*].fields[?(@.key == "jLIgoi")]').find(json)[0].value["answer"]
```
gidsg, line 85: This is an optional field, we need to coalesce it to 0 if not populated.

db/queries/assessment_records/_helpers.py
```diff
@@ -1,9 +1,12 @@
+import jsonpath_rw_ext
```
gidsg, line 1: We need to add this dependency to the requirements file.

I am getting this error:
```
ModuleNotFoundError: No module named 'jsonpath_rw_ext'
```

gidsg (approved):
👍 Nice fix.

db/queries/assessment_records/_helpers.py
```diff
@@ -1,79 +1,59 @@
...
+        asset_type = "Not asset type specified."
+    try:
+        funding_one = (
+            jsonpath_rw_ext.parse(
+                '$.forms[*].questions[*].fields[?(@.key =="JzWvhj")]'
+            )
+            .find(application_json)[0]
+            .value["answer"]
+        )
+    except Exception:
```
gidsg, line 31: Will this definitely throw an exception or will it return `None`?

Adam-W1: no in the event of a null in the json, it throws an error. I have changed this to set the value as 0 👍

PR #92, opened 2023-01-09
Bau fix config mutation

tferns:
In `get_all_subcriteria` - we were returning references to the original config.  This was then being used in a bunch of other logic via mutations, which were affecting the original config, I've changed it to use `copy.deepcopy` so we don't mutate the original refererences.

I also refactored the `SubCriteriaThemes` class to be pure functions on the module, since I originally thought this might've been the cause of the issue.


adamdavies1 (approved):
LGTM

PR #94, opened 2023-01-10
FS-2146: Fix latest sub criteria score query

tferns:
- Fix latest sub criteria score query limit
- Add test to verify changes function as intended

Previously, we were limiting to `1` - which was only returning a single score for the entire application id.
What we actually want, is the latest score for every unique sub criteria id in an assessment record.

### DIFF
--- a/db/queries/scores/queries.py
+++ b/db/queries/scores/queries.py
@@ -15,7 +15,7 @@ def get_scores_for_app_sub_crit(
     application_id: str, sub_criteria_id: str, score_history: bool = False
 ) -> list[dict]:
     """get_scores_for_app_sub_crit executes a query on scores
-    which returns the most recent score or all scores for the 
+    which returns the most recent score or all scores for the
     given application_id and sub_criteria_id.
 
     :param application_id: The stringified application UUID.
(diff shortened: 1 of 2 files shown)


db/queries/scores/queries.py
```diff
@@ -24,34 +24,45 @@ def get_scores_for_app_sub_crit(
...
-            Score.sub_criteria_id == sub_criteria_id
-            ).order_by(Score.date_created.desc())
+        stmt = (
+            select(Score)
+            .where(
+                Score.application_id == application_id,
+                Score.sub_criteria_id == sub_criteria_id,
+            )
+            .order_by(Score.date_created.desc())
+        )
```
gidsg, line 34: Could we move this part of the query into a variable and then just change the limit in the `else` case?

tferns: Ah I've actually not looked at this code - it was just pre-commit formatting since whoever had committed it didn't run.  I'll look at adding your suggestion

tferns: Refactored in https://github.com/communitiesuk/funding-service-design-assessment-store/pull/94/commits/0d51c337e6e310312c004c4ebd32774419efaafe

Adam-W1 (commented):
Looks good 👍 
can we add the expectation to see the scores when testing the assessment_dashboard route?

tferns:
> Looks good 👍 can we add the expectation to see the scores when testing the assessment_dashboard route?

Hey, I think that's logic relevant to the assessment frontend.  This PR is just for the back-end to pull the data back for the front-end.  Do you maybe mean something else and I'm misunderstanding?

Adam-W1:
> > Looks good 👍 can we add the expectation to see the scores when testing the assessment_dashboard route?
> 
> Hey, I think that's logic relevant to the assessment frontend. This PR is just for the back-end to pull the data back for the front-end. Do you maybe mean something else and I'm misunderstanding?

No worries, I did mean to add the test data for the frontend. However, I'm now adding some sample scores data in for the tests on fs-2018 so I think that will cover it 👍

Adam-W1 (approved):

PR #96, opened 2023-01-10
FS-1666 Add assessments stats endpoint

harryyo:
### Added assessments stats

- Added `/assessments/get-stats/{fund_id}/{round_id}` endpoint
- New flags_by_type query


tferns (approved):
Looks good, great work!

tferns:
> Looks good, great work!

Only comment is, I think the logic and code in this PR is sound and good.  But we had a small discussion about this model potentially changing so I'm not sure if that is being done in parallel by @AbuNavsa or @Adam-W1 - if so they may need to also update this endpoint to reflect the model changes.

api/routes/assessment_routes.py
```diff
@@ -114,3 +115,103 @@ def get_sub_criteria_theme_answers(application_id: str, theme_id: str):
...
+            ),
+    # TODO: add resolved if required, after flag_type enum
+    # has been updated to include QUERY_RESOLVED
+    #         "query_resolved": len(
+    #             [
+    #                 1
+    #                 for assessment in assessments
+    #                 if assessment["application_id"] in resolved_assessments
+    #             ]
+    #         ),
```
harryyo, line 212: Have just left this commented for now pending update of FlagType enum to contain QUERY_RESOLVED fyi @Adam-W1

harryyo: Removed

PR #98, opened 2023-01-11
Fs 2018 resolve a flagged application

Adam-W1:
https://digital.dclg.gov.uk/jira/browse/FS-2018
Resolving a flagged application
Linked to https://github.com/communitiesuk/funding-service-design-assessment/pull/131

**Changes to flags schema and flagging Enums. An example workflow using FLAGGED and STOPPED:**

Pasted Graphic 7


db/migrations/versions/3a22407701c8_.py
```diff
@@ -0,0 +1,42 @@
...
+revision = '3a22407701c8'
+down_revision = '99ad87a4f888'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+
+    op.drop_column("flags","resolution_reason")
```
gidsg, line 22: Think this is OK for now as no-one is using flags yet, but I think this would delete all data if we had some, so would need to come up with a more incremental migration in future.

gidsg (approved):
LGTM (as long as no-one is currently using flags on production?) 👍 🎏

PR #99, opened 2023-01-11
FS-2149: Show Flagged & fix filters

tferns:
- Show Flagged on assessor team dashboard
- Fix filters for Flagged applications, and unhandled statuses

See: https://github.com/communitiesuk/funding-service-design-assessment/pull/125

Video demo:
https://user-images.githubusercontent.com/117724519/211817671-47e51f30-bee5-4037-b822-81c1ffed6104.mp4

### DIFF
--- a/db/queries/assessment_records/queries.py
+++ b/db/queries/assessment_records/queries.py
@@ -38,10 +38,6 @@ def get_metadata_for_fund_round_id(
     :return: A list of dictionaries.
     """
 
-    # TODO: Handle these statuses
-    if status in ("INELIGIBLE", "QA_COMPLETE", "QA_READY"):
-        return []
-
     statement = (
         select(AssessmentRecord)
         # Dont load json into memory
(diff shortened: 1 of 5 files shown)


db/queries/flags/queries.py
```diff
@@ -33,9 +33,9 @@ def create_flag_for_application(
     return flag_metadata
 
 
-def retrieve_flags_for_application(application_id: str) -> dict:
+def retrieve_flags_for_application(application_ids: list[str]) -> dict:
```
harryyo, line 36: Just minor, but could this be renamed to retrieve_flags_for_application**s**() so that it's a bit more clear that it's a bulk function now?

harryyo (approved):
Looks good, just one comment

db/queries/assessment_records/queries.py
```diff
@@ -57,11 +61,31 @@ def get_metadata_for_fund_round_id(
     if asset_type != "ALL" and asset_type != "":
         statement = statement.where(AssessmentRecord.asset_type == asset_type)
 
-    if status != "ALL" and status != "":
+    if status not in ("", "ALL", "FLAGGED"):
```
gidsg, line 64: Could we do a pattern match on status rather than a series of if statements i.e. https://peps.python.org/pep-0634/?

db/queries/flags/queries.py
tferns: Good spot have added in https://github.com/communitiesuk/funding-service-design-assessment-store/pull/99/commits/d7740ce61e9a0f6a93cf9fe5681f66fc13b41016

gidsg (approved):

PR #100, opened 2023-01-12
BAU: Change progress endpoint to post

tferns:
- Start posting data as a json body as opposed to query params

https://sentry.io/organizations/funding-service-design-team-dl/issues/3850987122/?referrer=slack


tferns:
> LGTM, might be worth in future leaving the backwards compatible route in temporarily so the frontend/backend deploys don't need to be syncronised.

Good point I've added it back in https://github.com/communitiesuk/funding-service-design-assessment-store/pull/100/commits/70ca76e573ba65cc8b7a79f484a496327c8dfcc4

PR #102, opened 2023-01-16
fix migration to change enum rather than replace the column

Adam-W1:
Realtes to https://digital.dclg.gov.uk/jira/browse/FS-2018 failed build (https://github.com/communitiesuk/funding-service-design-assessment-store/actions/runs/3928462907/jobs/6716156366).

Fix the migration to change enum rather than replace the column


db/migrations/versions/3a22407701c8_.py
```diff
@@ -18,25 +18,24 @@
 
 def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
-
+    with op.get_context().autocommit_block():
+        op.execute("ALTER TYPE flagtype ADD VALUE 'RESOLVED'")
     op.drop_column("flags","resolution_reason")
     op.execute("DROP TYPE resolutiontype;")
```
gidsg, line 24: LGTM, as long as no-one is using these yet.

gidsg (approved):
👍

db/migrations/versions/3a22407701c8_.py
Adam-W1: OKay great, ive tested with resolution data on downgrades and upgrades and was happy so hopefully this wont cause issues either way (the column is nullable)

PR #58, opened 2022-12-08
FS-1987-show-sub-criteria-themes-to-assessors

ramsharma-prog:

PR #104, opened 2023-01-17
FS-2188: Change to use rsplit with limit of 1

tferns:
We're splitting on `: ` from the left, without any limit.  The user has entered a `:`  as part of their description, so we're splitting pre-maturely and incorrectly, which subsequently messes up the float coercion on the frontend. Changing to `rsplit` with a limit of `1`, should fix the immediate issue.

Relevant links (for reference):
- https://github.com/communitiesuk/funding-service-design-assessment-store/pull/58#discussion_r1047347313
- [slack thread](https://communities-govuk.slack.com/archives/C02PHBER287/p1673882389522519)
- [JIRA FS-2188](https://digital.dclg.gov.uk/jira/browse/FS-2188)

### DIFF
--- a/tests/test_get_sub_criteria.py
+++ b/tests/test_get_sub_criteria.py
@@ -20,6 +20,11 @@ from api.routes.subcriterias.get_sub_criteria import (
             ["£1"],
         ),
         (
+            ["An : Answer : with : lots: of :colons : £1"],
+            ["An : Answer : with : lots: of :colons "],
+            ["£1"],
+        ),
+        (
             ["An Answer : £1", "Another Answer : £2"],
             ["An Answer ", "Another Answer "],
             ["£1", "£2"],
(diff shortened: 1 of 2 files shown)


tests/test_get_sub_criteria.py
```diff
@@ -1,11 +1,36 @@
...
+            ["An Answer "],
+            ["£1"],
+        ),
+        (
+            ["An Answer : with colon : £1"],
+            ["An Answer : with colon "],
+            ["£1"],
+        ),
+        (
+            ["An Answer : £1", "Another Answer : £2"],
```
gidsg, line 23: Can we add a test with multiple `:` characters in the description part? We should be able to handle any number and only the last one should be the amount.

tferns: This is being done just above on [L18](https://github.com/communitiesuk/funding-service-design-assessment-store/blob/023feaf7c3ff9f76bf157ec40f157dd5827b35aa/tests/test_get_sub_criteria.py#L18) - or do you mean even more colons?

gidsg (approved):
LGTM 👍 🐛

PR #105, opened 2023-01-17
FS-2163 - Populate location data in assessment-store

srh-sloan:
Covers the following:
- Creating a script (`scripts/populate_location_data.py`) to extract postcode from `address of asset` field, then retrieve more detailed location data and populate the `location_json_blob` field in the DB.
- Updating test data functions accordingly (ie being able to seed the dev/unit_test DBs with this location data)
- Also added option to manually run the workflow and deploy to dev. This is integrated in the main `.github/workflows/deploy.yml` file, not a separate 'deploy_to_dev' workflow to reduce duplication.


README.md
```diff
@@ -85,6 +85,17 @@ upgrade the deployed db instances with your changes.
 
 ## How to use
 Enter the virtual environment and setup the db as described above, then:
+## How to run locally
+Enter the virtual environment and install dependencies as described above, then:
+
+### Create and seed local DB
+- Make sure your local `DB_URL` env var is set to your local postgres db (this doesn't need to actually exist yet), eg:
```
gidsg, line 92: Should this be `DATABASE_URL`?

db/queries/assessment_records/queries.py
```diff
@@ -270,3 +272,22 @@ def get_application_jsonb_blob(application_id: str) -> dict:
...
+    stmt = (
+        update(AssessmentRecord)
+        .where(AssessmentRecord.application_id == bindparam("app_id"))
+        .values(location_json_blob=bindparam("new_location_data"))
+    )
+
+    update_params = [
+        {
+            "app_id": item["application_id"],
+            "new_location_data": item["location"],
```
gidsg, line 287: I guess it's a bit late as we've done the migration already? But can we avoid temporal references if possible? So `location_data` would be preferable to `new_location_data`.

README.md
srh-sloan: Yes, thanks!

scripts/import_from_application.py
```diff
@@ -1,11 +1,16 @@
...
 import argparse
+import sys
 
 import requests
-from app import app
-from config import Config
-from db.queries import bulk_insert_application_record
-from fsd_utils import CommonConfig
+
+sys.path.insert(1, ".")
```
gidsg, line 7: Could we try using `python -m` flag with path to the script as per https://github.com/communitiesuk/funding-service-design-assessment-store/blob/066dec1ea4003a01a2c15d17f86884959f71e88d/scripts/README.md?plain=1#L6 to try and avoid this workaround?

README.md
```diff
@@ -121,12 +132,18 @@ These are the current pipelines running on the repo:
 
 # Testing
 
+The tests within /tests are designed to connect to a real database to run tests that use the DB layer, details below. This takes a long time to setup on a test run, so if your tests don't need a DB put them in /tests_no_db. This folder doesn't pickup /tests/conftest.py so therefore doesn't do the DB setup work and these tests can run in isolation much faster.
+
 ## Seed Postgres DB with mock data
 
 `invoke seed_dev_db`
 
 Running the above command will prompt you to enter the number applications, funds & rounds you would like to create as mock data within the database.
```
gidsg, line 141: typo 'number _of_ applications'.

README.md
```diff
@@ -121,12 +132,18 @@ These are the current pipelines running on the repo:
...
 
+This will also work for the DB within the docker runner. Find the ID of the docker container running assessment-store (`docker ps`) then execute:
+
+        docker exec -it <container_id> invoke seed_dev_db
```
gidsg, line 145: I think it's a bit confusing the mix between the docker and local instructions but I think maybe we can pick this up as part of a bigger bit of work looking at consistent use of databases and how to run the tests, so outside the scopr of this ticket.

scripts/import_from_application.py
srh-sloan: Yeah that works 👍

scripts/README.md
```diff
@@ -4,3 +4,26 @@
...
+1. Updates the database, populating `assessment_records.location_json_blob` with a json blob containing the location details
+1. Writes the retrieved location data out to a csv file at the specified location.
+
+To run locally:
+
+        ./scripts/populate_location_data.py --fund_id 47aef2f5-3fcb-4d45-acb5-f0152b5f03c4 --round_id c603d114-5364-4474-a0c4-c41cbf4d3bbd --update_db True --write_csv True --csv_location ./locations.csv
+
+To run on paas, execute the following (starts a task using the app context etc)
+
+        cf run-task funding-service-design-assessment-store-dev --command "./scripts/populate_location_data.py --fund_id 47aef2f5-3fcb-4d45-acb5-f0152b5f03c4 --round_id c603d114-5364-4474-a0c4-c41cbf4d3bbd --update_db True --write_csv True --csv_location /tmp/locations.csv && cat /tmp/locations.csv" --name populate_location
```
gidsg, line 23: Will these commands use `python -m` now? If so please can we update?

srh-sloan: Yes, good shout

scripts/populate_location_data.py
```diff
@@ -46,7 +43,7 @@ def process_locations(
             postcode = get_postcode_from_questions(questions)
             application_ids_to_postcodes[id] = postcode
             just_postcodes.append(postcode)
-
+        print(just_postcodes)
```
gidsg, line 46: Do we need to print these or is that just for debugging?

srh-sloan: 🤦‍♀️

gidsg (approved):
LGTM 👍 I would maybe suggest separating the commands using `cf ssh` for the CSV command so we could get the output without having to copy it out the logs, but OK as-is if we don't want to do that.

Please squash and merge as commit history is quite messy.

srh-sloan:
> LGTM 👍 I would maybe suggest separating the commands using `cf ssh` for the CSV command so we could get the output without having to copy it out the logs, but OK as-is if we don't want to do that.
> 
> Please squash and merge as commit history is quite messy.

I don't think we can separate out into a `cf ssh` call as the container to run the task is spun up and then destroyed when the task ends.

scripts/README.md
```diff
@@ -4,3 +4,26 @@
...
 `python -m scripts.import_from_application --roundid={your round id}`
+
+## Populate Location Data
+
+This script does the following:
+
+1. Extracts postcode from the 'address of asset' field in the application form
+1. Sends these postcodes to postcodes.io to retrieve further location information, eg. country
+1. Updates the database, populating `assessment_records.location_json_blob` with a json blob containing the location details
+1. Writes the retrieved location data out to a csv file at the specified location.
```
Adam-W1, line 15: think this should be incrementing. Lovely docs though 👍

srh-sloan: That happens anyway - you just use `1.` every time and it increments them when it renders. Saves having to renumber everything if you add a new bullet mid-list :)

Adam-W1: ah very fancy, ill have to remember that one 👌

PR #106, opened 2023-01-18
fs-1524-mark-qa-complete

gio-karunakaran:


db/queries/assessment_records/queries.py
```diff
@@ -283,4 +283,14 @@ def update_status_to_completed(application_id):
         {AssessmentRecord.workflow_status: Status.COMPLETED}, synchronize_session=False)
 
     db.session.commit()
-    
\ No newline at end of file
+
+def update_status_to_qa_complete(application_id):
```
gidsg, line 287: Can we combine this with the function above e.g. `update_status` and pass the status enum in as an argument?

db/migrations/versions/f586493c6c55_.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute("ALTER TYPE status RENAME TO status_old")
+    op.execute("CREATE TYPE status AS ENUM('NOT_STARTED', 'IN_PROGRESS', 'SUBMITTED', 'COMPLETED')")
+    op.execute((
+        "ALTER TABLE transactions ALTER COLUMN status TYPE status USING "
```
gidsg, line 33: our table name is not `transactions`. Please can we use the correct table name?

db/queries/assessment_records/queries.py
gio-karunakaran: done!

db/migrations/versions/f586493c6c55_.py
gio-karunakaran: sorry missed that! have changed this now 👍🏾

db/queries/assessment_records/queries.py
```diff
@@ -273,14 +274,20 @@ def get_application_jsonb_blob(application_id: str) -> dict:
...
 
-def update_status_to_completed(application_id):
-    from flask import current_app
-    current_app.logger.info(
-             f"Updating application status to COMPLETED")
-    db.session.query(AssessmentRecord)\
+def update_status(application_id, status):
+    if status == "COMPLETED":
+        current_app.logger.info("Updating application status to COMPLETED")
+        db.session.query(AssessmentRecord)\
```
gidsg, line 280: We are making very poor use of the argument here. Could we have a single query and pass in the status to it?

We can lookup the status from String using array syntax e.g. `Status['COMPLETED']`

If we want to constrain the input we could check at the start of the function e.g.
```
if status not in ['COMPLETED', 'QA_COMPLETE']:
    abort(400, "Incorrect Status given")
```

db/queries/assessment_records/queries.py
```diff
@@ -273,14 +274,13 @@ def get_application_jsonb_blob(application_id: str) -> dict:
     return application_json
 
 
-def update_status_to_completed(application_id):
-    from flask import current_app
-    current_app.logger.info(
-             f"Updating application status to COMPLETED")
+def update_status(application_id, status):
+    if status not in Status.__members__:
```
gidsg, line 278: Do we want to allow all statuses here? Previously we were just allowing `QA_COMPLETE` or `COMPLETED`?

tests/test_routes.py
```diff
@@ -141,15 +141,15 @@ def test_get_sub_criteria_theme_answers_field_id(request, client):
     assert response.json[0]["field_id"] == "ieRCkI"
 
 
-def test_update_ar_status_to_completed(request, client):
+def test_update_ar_status(request, client):
```
gidsg, line 144: Should we test with multiple statuses?
Can we test that the status has actually updated on the application?

db/queries/assessment_records/queries.py
gio-karunakaran: thought it could be worth allowing all statuses to future-proof this function/ make more flexible if needed?

tests/test_routes.py
gio-karunakaran: now testing with both COMPLETED and QA_COMPLETE statuses, and checking db rows are updated 👍🏾

api/routes/assessment_routes.py
```diff
@@ -119,9 +119,9 @@ def get_sub_criteria_theme_answers(application_id: str, theme_id: str):
     return map_application_with_sub_criteria_themes(application_id, theme_id)
 
 
-def update_ar_status_to_completed(application_id: str):
-    """ Function updates the status to COMPLETE for the given application_id"""
-    update_status_to_completed(application_id)
+def update_ar_status(application_id: str, status: str):
```
gidsg, line 122: Can we try to avoid acronyms, so in this case can we call it `update_assessment_status`?

db/queries/assessment_records/queries.py
gio-karunakaran: done!

gio-karunakaran:
closing as easier to re-start on a fresh branch

PR #107, opened 2023-01-18
Fs 2163 csv fields

srh-sloan:
Updated CSV writing to use the application short ID in order to match what COF use in their tracker.
Also added default values for location fields if they are unavailable (eg. dodgy postcode supplied) as per design discussions


db/queries/assessment_records/_helpers.py
```diff
@@ -61,5 +61,17 @@ def derive_application_values(application_json):
...
+
+    FIELD_DEFAULT_VALUE = "Not Available"
+    if derived_values["location_json_blob"]["error"] is True:
+        derived_values["location_json_blob"]["county"] = FIELD_DEFAULT_VALUE
+        derived_values["location_json_blob"]["region"] = FIELD_DEFAULT_VALUE
+        derived_values["location_json_blob"]["country"] = FIELD_DEFAULT_VALUE
+        derived_values["location_json_blob"][
+            "constituency"
+        ] = FIELD_DEFAULT_VALUE
+        derived_values["location_json_blob"]["postcode"] = FIELD_DEFAULT_VALUE
```
gidsg, line 75: Looks like this is in the common import script. Is the location data script separate from the bulk import functionality, or am I mis-understanding?

srh-sloan: Yes the location script is separate from the bulk import from application script, as the location one wasn't done when we first ran the bulk import in live.
But the scripts we use to generate test data also use this method (eg. when running `invoke seed_dev_db`) so the script needs to account for it being there or not, hence the `if` check on line 60.
As it stands, when this goes live the location will still show as blank until we run the script, but for any future funds or for test data it will show as 'Not Available'

db/queries/assessment_records/_helpers.py
```diff
@@ -61,5 +61,17 @@ def derive_application_values(application_json):
...
             "location_json_blob"
         ]
+    else:
+        derived_values["location_json_blob"] = {"error": True}
+
+    FIELD_DEFAULT_VALUE = "Not Available"
+    if derived_values["location_json_blob"]["error"] is True:
+        derived_values["location_json_blob"]["county"] = FIELD_DEFAULT_VALUE
+        derived_values["location_json_blob"]["region"] = FIELD_DEFAULT_VALUE
+        derived_values["location_json_blob"]["country"] = FIELD_DEFAULT_VALUE
```
gidsg, line 71: I would prefer leaving these as `None` as this label is presentational and should be controlled in the frontend rather than set in the database.

johnoneill46 (approved):

PR #108, opened 2023-01-19
fs-1524-flag-qa-complete

gio-karunakaran:


tferns (approved):
Change look good, as long as existing data isn't corrupted in any way by the migration? 

Don't think it will be a problem, but wanted to mention just incase.

PR #112, opened 2023-01-20
Nest flags in overview

AbuNavsa:
Nested flags into application_overview schema to retrieve flags with the application, for use on dashboard to display STOPPED in status if STOPPED is the most recent

### DIFF
--- a/db/schemas/schemas.py
+++ b/db/schemas/schemas.py
@@ -10,6 +10,7 @@ from marshmallow import fields
 from marshmallow import Schema
 from marshmallow.fields import Enum
 from marshmallow.fields import Field
+from marshmallow.fields import Nested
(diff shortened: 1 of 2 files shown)


db/schemas/schemas.py
```diff
@@ -23,7 +24,7 @@ class Meta:
 
     workflow_status = Enum(Status)
     language = Enum(Language)
-
+    flags = Nested("FlagMetadata", many=True)
```
gidsg, line 27: Is it possible to add a test for this to understand how it is affecting the JSON response?

AbuNavsa: Yes, I will add testing

gidsg (approved):

PR #114, opened 2023-01-24
fs-1524-qa-complete-flag

gio-karunakaran:


Adam-W1 (commented):
looks good :) you might need to add 'is_qa_complete' to the components.yml Flag response for when APIs are talking

Adam-W1 (commented):
maybe a test for this also to get coverage

db/queries/flags/queries.py
```diff
@@ -47,6 +47,18 @@ def retrieve_flag_for_application(application_id: str) -> dict:
...
+def find_qa_complete_flag(application_id):
+    qa_complete_flag = (
+        Flag.query.filter(
+            Flag.application_id == application_id,
+            Flag.flag_type == "QA_COMPLETED"
+        )
+        .first()
+    )
+
+    return {"is_qa_complete": True if qa_complete_flag else False}
```
gidsg, line 59: can simplify logic
```suggestion
    return {"is_qa_complete": bool(qa_complete_flag)}
```

gio-karunakaran: love it!

PR #118, opened 2023-02-06
FS-833: Added SQL query for COF assessment reporting

johnoneill46:
https://digital.dclg.gov.uk/jira/browse/FS-833

### DIFF
--- a/scripts/assessment-cof-reporting.sql
+++ b/scripts/assessment-cof-reporting.sql
@@ -1,0 +1,24 @@
+SELECT
+	application_id,
+	short_id,
+	jsonb_path_query_first (jsonb_blob, '$.forms[*].questions[*].fields[*] ? (@.key == "WWWWxy")."answer"') AS "Unique Tracker Number",
+	jsonb_path_query_first (jsonb_blob, '$.forms[*].questions[*].fields[*] ? (@.key == "YdtlQZ")."answer"') AS "Organisation Name",
+	jsonb_path_query_first (jsonb_blob, '$.forms[*].questions[*].fields[*] ? (@.key == "lajFtB")."answer"') AS "Organisation Type",
+	jsonb_path_query_first (jsonb_blob, '$.forms[*].questions[*].fields[*] ? (@.key == "KAgrBz")."answer"') AS "Project Name",
+	jsonb_path_query_first (jsonb_blob, '$.forms[*].questions[*].fields[*] ? (@.key == "LZBrEj")."answer"') AS "Lead Contact Email Address",
+	funding_amount_requested AS "Funding Amount Requested",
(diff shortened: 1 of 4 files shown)


scripts/assessment-cof-reporting.sql
```diff
@@ -0,0 +1,24 @@
...
+	jsonb_path_query_first (jsonb_blob, '$.forms[*].questions[*].fields[*] ? (@.key == "NyudvF")."answer"') AS "Revenue Costs",
+	asset_type AS "Asset Type",
+	jsonb_path_query_first (jsonb_blob, '$.forms[*].questions[*].fields[*] ? (@.key == "GjzaqR")."answer"') AS "Asset Type Other",
+
+	jsonb_path_query_first (jsonb_blob, '$.forms[*].questions[*].fields[*] ? (@.key == "yEmHpp")."answer"') AS "Address",
+
+	location_json_blob -> 'postcode' AS "Postcode",
+	location_json_blob -> 'county' AS "County",
+	location_json_blob -> 'country' AS "Country",
+	location_json_blob -> 'constituency' AS "Constituency"
```
gidsg, line 20: This doesn't include any scores or comments, are we planning on adding those or doing them in separate queries?

gidsg: Can we convert to a draft PR if we're still amending it?

johnoneill46: Scores and comments will be pulled in as separate queries but these are simple queries. I wasn't planning on adding them to github as they're so simple however

gidsg: If there is some sort of manual recipe we're using to pull the final CSV together, can we still add these queries and document the process as-is? Ideally we should be able to pull all the data in a single query but if not at least document the process as-is.

johnoneill46: Other scripts included. Once I retrieve the data I'll merge it together as requested by Martin manually

scripts/assessment-cof-reporting_scores.sql
```diff
@@ -0,0 +1,6 @@
+SELECT
+	sub_criteria_id AS "Score Subcriteria",
+	score AS "Score",
+	justification AS "Score Justification"
+FROM
+	scores
```
gidsg, line 6: As I understand it this is just getting all the scores? How are they linked to the application ID? Also subcriterias can be 're-scored' so we need to ensure we're only taking the latest in each case.

scripts/assessment-cof-reporting_scores.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+	application_id AS "Application ID",
+	sub_criteria_id AS "Score Subcriteria",
+	score AS "Score",
```
gidsg, line 4: This is incorrect I believe as we only want the latest score per subcriteria else we risk including scores which have since been rescored.

PR #121, opened 2023-02-07
FS-2289 - Fixed metric issue, now uses retrieve flags endpoint

AbuNavsa:
### Issue:
The reason for the inaccurate metric is because the endpoint retrieving the applications data retrieves the following metrics for example: 

{'assessing': 0, 'completed': 3, 'flagged': 0, 'not_started': 7, 'qa_completed': 3, 'stopped': 0, 'total': 10}
 
At this point the metric will show 3 QA Completed, however, if we now FLAG an application, although it is still QA completed, since the latest flag is "Flagged", the following will show:

{'assessing': 0, 'completed': 3, 'flagged': 1, 'not_started': 7, 'qa_completed': 2, 'stopped': 0, 'total': 10}

The reason for this is because the endpoint retrieving 'qa_completed' will only count it as true if that is the latest flag. 

### Fix:

The query used to retrieve the amount of qa_completed applications previously only returned true IF qa_completed was the latest flag. I have changed this to use the new query find_qa_complete_flags_for_applications query which searches for those applications which have a qa completed flag in them

### DIFF
--- a/api/routes/assessment_routes.py
+++ b/api/routes/assessment_routes.py
@@ -14,6 +14,7 @@ from db.queries import get_metadata_for_fund_round_id
 from db.queries import retrieve_flags_for_applications
 from db.queries.assessment_records.queries import find_assessor_task_list_state, update_status_to_completed
 from db.queries.flags.queries import get_latest_flags_for_each
+from db.queries.flags.queries import find_qa_complete_flag_for_applications
 from db.queries.assessment_records.queries import find_assessor_task_list_state
 from db.queries.assessment_records.queries import ( 
     get_assessment_sub_critera_state,
@@ -141,14 +142,13 @@ def assessment_stats_for_fund_round_id(
     stats = {}
     assessments = get_metadata_for_fund_round_id(
         fund_id=fund_id, round_id=round_id
-    )
+    )s
     assessment_ids = [application['application_id'] for application in assessments ]
 
-    qa_completed_assessments = [
+    qa_completed_assessments = [    
         flag["application_id"]
-        for flag in retrieve_flags_for_applications(assessment_ids)
-    ]
-    
+        for flag in find_qa_complete_flag_for_applications(assessment_ids)
+    ]    
     stopped_assessments = [
         flag["application_id"] for flag in get_latest_flags_for_each("STOPPED")
     ]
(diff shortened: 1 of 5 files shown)


gidsg:
Please include the ticket number in the PR title.

api/routes/assessment_routes.py
```diff
@@ -141,10 +142,13 @@ def assessment_stats_for_fund_round_id(
     assessments = get_metadata_for_fund_round_id(
         fund_id=fund_id, round_id=round_id
     )
+    assessment_ids = [application['application_id'] for application in assessments ]
+
     qa_completed_assessments = [
         flag["application_id"]
-        for flag in get_latest_flags_for_each("QA_COMPLETED")
+        for flag in retrieve_flags_for_applications(assessment_ids)
```
gidsg, line 149: Don't think this is actually filtering on `QA_COMPLETED` anymore. it would just return all flags?

api/routes/assessment_routes.py
```diff
@@ -141,14 +142,13 @@ def assessment_stats_for_fund_round_id(
     stats = {}
     assessments = get_metadata_for_fund_round_id(
         fund_id=fund_id, round_id=round_id
-    )
+    )s
```
gidsg, line 145: Looks like this stray `s` is causing the tests to fail. Have you got the pre-commit hooks set-up?

gidsg:
Can we try adding a test for this logic?

api/routes/assessment_routes.py
AbuNavsa: It filters later on however, have changed to use a query just looking for applications of QA_COMPLETED now too

api/routes/assessment_routes.py
AbuNavsa: Removed and fixed, thanks!

tests/test_db_function.py
```diff
@@ -414,6 +415,26 @@ def test_retrieve_flag_for_application(db_session):
     assert result["user_id"] == second_flag.user_id
     assert result["flag_type"] == second_flag.flag_type.name
 
+def test_find_qa_complete_flag_for_applications(db_session):
+    """Put QA_COMPLETED flags in 1 out of 2 applications and
+    only retrieve the metadata for the 1 with QA_COMPLETED"""
+
+    first_flag = Flag(**flag_config[2])
```
gidsg, line 422: Can we use variable names for these flags with the flag type e.g. `qa_complete_flag` rather than `first_flag` etc. else it's a bit opaque what's going on?

gidsg:
Can we try adding a test for the actual endpoint response? i.e. using the flask test client?

gidsg (approved):
👍

PR #123, opened 2023-02-09
FS-2294: return all comments if theme_id not supplied

gidsg:
Previously we returned all comments only if a parameter of `theme_id=score` was supplied.

This is not semantically meaningful and was presumably done to match the similarly confusing behaviour in the frontend.

This changes the behaviour to return all comments for a subcriteria if no theme is supplied. It maintains the old behaviour for backwards compatibility until the frontend is deployed without using `theme_id=score`


Adam-W1 (approved):
👍

PR #165, opened 2023-04-26
update seed data and add a patch fix

Adam-W1:
_Add ticket reference to Pull Request title: e.g. 'FS-123: Add content', if there is no ticket prefix with BAU_

### Change description
https://digital.dclg.gov.uk/jira/browse/FS-2500

Currently, the add-another component output is not consumable (going forward) by the assessment service, this is because the keys are dictated by the field name in the form builder. This form builder context is not known by assessment when extracting the information (by key) from the application payload. Instead we should use consistent keys such as "name" and "value" as below.

For now a patch has been put in place to handle the temporary "add-another" output.


tferns:
Superseded by https://github.com/communitiesuk/funding-service-design-assessment-store/pull/172

PR #166, opened 2023-04-28
FS-2640-location-missing-for-some-applications

ramsharma-prog:
This PR  returns a dictionary with default values indicating that a location was not found.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -27,4 +27,4 @@ The `cat /tmp/locations.csv` prints out the csv to the logs so we can copy/paste
 
 If there is data in your docker DB, you can also execute this script locally in the container:
 
-        docker exec -ti bb326fba2d6c scripts/populate_location_data.py --fund_id 47aef2f5-3fcb-4d45-acb5-f0152b5f03c4 --round_id c603d114-5364-4474-a0c4-c41cbf4d3bbd --update_db True --write_csv False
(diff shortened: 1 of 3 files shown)


rayhan15243 (approved):

scripts/README.md
```diff
@@ -27,4 +27,4 @@ The `cat /tmp/locations.csv` prints out the csv to the logs so we can copy/paste
 
 If there is data in your docker DB, you can also execute this script locally in the container:
 
-        docker exec -it db6f6f6a2ee0 ./scripts/populate_location_data.py --fund_id 47aef2f5-3fcb-4d45-acb5-f0152b5f03c4 --round_id c603d114-5364-4474-a0c4-c41cbf4d3bbd --update_db True --write_csv False
+        docker exec -ti db6f6f6a2ee0 scripts/populate_location_data.py --fund_id 47aef2f5-3fcb-4d45-acb5-f0152b5f03c4 --round_id c603d114-5364-4474-a0c4-c41cbf4d3bbd --update_db True --write_csv False
```
gio-karunakaran, line 30: should we keep both these lines as there are scenarios where one will work and not the other? depending on machine/shell being used?

gio-karunakaran (commented):
Looks good, just a comment

scripts/README.md
ramsharma-prog: Are you referring it to -it and -ti or the removing ./ ?

ramsharma-prog: adding ./ sometimes doesn't import all the files so with adding sys.path.insert(1, ".") makes sure the script goes to the home directory

gio-karunakaran: ahhh perfect that makes sense :)

gio-karunakaran (approved):
LGTM

PR #168, opened 2023-05-04
FS-2640-location-is-missing-for-some-applications

ramsharma-prog:
This PR adds a condition that will result in a "Not found" response if the Project information page for a community asset address cannot be found. This behavior will only occur in local environments since the Project information address field is required.

### DIFF
--- a/scripts/location_utils.py
+++ b/scripts/location_utils.py
@@ -36,6 +36,12 @@ def get_postcode_from_questions(form_questions) -> str:
                     postcode = raw_postcode.strip().replace(" ", "").upper()
                     return postcode
 
+            else:
+                print(
+                    "The key 'yEmHpp' for Project information form not found"
+                )
+                return "Not found"
+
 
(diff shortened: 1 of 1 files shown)


scripts/location_utils.py
```diff
@@ -36,6 +36,12 @@ def get_postcode_from_questions(form_questions) -> str:
                     postcode = raw_postcode.strip().replace(" ", "").upper()
                     return postcode
 
+            else:
+                print(
+                    "The key 'yEmHpp' for Project information form not found"
+                )
+                return "Not found"
```
RamuniN, line 43: Hopefully there is no postcode named "Not found" 😉

RamuniN (approved):
LGTM

scripts/location_utils.py
ramsharma-prog: Hope not 😆 but it shouldn't happen in production anyway

PR #173, opened 2023-05-10
FS-2689: Change maximum line length to 120

tferns:
### Change description

- Add flake8 argument for 120 line length
- Remove black argument for line length


gio-karunakaran (approved):
👍🏾

PR #197, opened 2023-06-22
Fs 3022 generalise populate location data for other funds

RamuniN:
https://digital.dclg.gov.uk/jira/browse/FS-3022
https://digital.dclg.gov.uk/jira/browse/FS-3023

### Description
- Refactored `get_postcode_from_questions` function to work with other funds
- Implemented location extraction in the `derive_application_values` function part of import script(so for future imports no need to run `populate_location_data` script)
- Make location data consistent to `Not Available` when not found
- Improved the import & populate script working with shortnames (see below How to test section)
- Made import script resilient by logging the failed application_id and continuing the import ([FS-3023](https://digital.dclg.gov.uk/jira/browse/FS-3023))
- Updated read me file

- [x] README and other documentation has been updated / added (if needed)
- [x] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test

#### To test import script
`docker exec -it <container_id> python -m scripts.import_from_application --roundid=c603d114-5364-4474-a0c4-c41cbf4d3bbd --app_type=COF`
or with short name
`docker exec -it <container_id> python -m scripts.import_from_application --fundround=COFR3W1`

#### To test populate location data script
`docker exec -ti <container_id> scripts/populate_location_data.py --fund_id 47aef2f5-3fcb-4d45-acb5-f0152b5f03c4 --round_id c603d114-5364-4474-a0c4-c41cbf4d3bbd --update_db True --write_csv False`
or with short name
`docker exec -ti <container_id> scripts/populate_location_data.py --fundround COFR2W3 --update_db True --write_csv False`

Alternatively, if using VsCode select the relevant debugger to test( see [here](https://github.com/communitiesuk/funding-service-design-assessment-store#launch-configurations-in-vscode))

### DIFF
--- a/db/queries/assessment_records/_helpers.py
+++ b/db/queries/assessment_records/_helpers.py
@@ -1,5 +1,8 @@
 import jsonpath_rw_ext
 import requests
+from config.mappings.assessment_mapping_fund_round import (
+    fund_round_data_key_mappings,
+)
 
 
 def get_answer_value(application_json, answer_key):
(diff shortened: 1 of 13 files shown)


db/queries/assessment_records/_helpers.py
```diff
@@ -61,6 +62,22 @@ def derive_application_values(application_json):
             + f"{application_id}."
         )
         funding_two = 0
+    try:
+        if "COF-R2" in short_ref:  # cof r2
```
srh-sloan, line 66: Shouldn't we be pulling these refs from the mapping above in assessment_mapping_fund_round?

db/queries/assessment_records/_helpers.py
srh-sloan, line 1: Line 28 should say 'No asset type' rather than 'Not asset type'

db/queries/assessment_records/_helpers.py
```diff
@@ -75,7 +92,32 @@ def derive_application_values(application_json):
             "location_json_blob"
         ]
     else:
-        derived_values["location_json_blob"] = {"error": True}
+        try:
```
srh-sloan, line 95: This whole bit is making `derive_application_values` a really massive, hard to read function. Could we pre-fetch this and pass it in, or separate out the postcodes.io bit into a separate function - easier to read and test.

db/queries/assessment_records/_helpers.py
srh-sloan: ACtually instead of all these if statements, can we just extend the stuff in [config/mappings/assessment_mapping_fund_round.py](https://github.com/communitiesuk/funding-service-design-assessment-store/pull/197/files#diff-5766d9b1f45c9da1bd0580dbbb2e8c58d9c40ca895b9e5ac05a60601bc57200d) to something more like:
`{"NSTFR2": {
    "location": "xyxyxy",
    "asset_type": None,
    "funding_one": "abcabc",
    "funding_two": "werwer" },
"COFR3W1": {
   "location": "oiuoiu",
etc...
}
}`

Then just grab stuff out of there for each specific field we need to populate

scripts/location_utils.py
```diff
@@ -21,16 +24,17 @@ def get_application_form(app_json_blob):
     ]
 
 
-def get_postcode_from_questions(form_questions) -> str:
+def get_postcode_from_questions(form_questions, fundround) -> str:
```
srh-sloan, line 27: This method is now duplicating what you've got in `derive_application_values` - can we just have one implementation?

srh-sloan (changes requested):
A few comments on de-duplicating and simplifying stuff. Also have we got tests for import_from_application? I know you've updated the location_utils test but we should create some (if they don't already exist) for import_from_application.

scripts/location_utils.py
RamuniN: It's not duplicating. In `derive_application_values` it is using entire json and here form_questions is only part of application which is also not a json.  so can't reuse  the code

db/queries/assessment_records/_helpers.py
RamuniN: Done

db/queries/assessment_records/_helpers.py
RamuniN: Done

srh-sloan (approved):
LGTM 👍

PR #208, opened 2023-07-04
fix get_flag_v2 endpoint

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3057

### Description
- part of the work for FS-3057
- fixed get flag v2 endpoint


ramsharma-prog (approved):
LGTM

PR #221, opened 2023-07-10
fs-3056-refactor-QA-complete

gio-karunakaran:
Changing QA complete from a flag to a status, adding a QA complete table that has a foreign key relationship to the assessment record table so that we can appropriately store the user and timestamp info of when an application was marked as QA complete. This PR also includes a lot of the re-work and logic changes to how qa complete is implemented as it changes from being a flag to a status.


db/models/flags_v2/flag_update.py
```diff
@@ -16,8 +16,7 @@
 class FlagStatus(IntEnum):
     RAISED = 0
     STOPPED = 1
-    QA_COMPLETED = 2
-    RESOLVED = 3
+    RESOLVED = 2
```
RamuniN, line 19: we need to be careful with `RESOLVED = 2` change. Assessment has to be updated to reflect this change

gio-karunakaran: ahh yes I wonder how this will affect existing records in the db?

RamuniN:
Looks good so far. I think you also need to create endpoint similar to [here ](https://github.com/communitiesuk/funding-service-design-assessment-store/blob/15ad8c0e047d03ff9438e8a5b11c584ea0837655/openapi/api.yml#L505)to insert QA Complete status.

db/models/flags_v2/flag_update.py
RamuniN: I think in the db, enum name is stored not the enum value so no effect of changing `RESOLVED = 2` to on existing records. But for removing `QA_COMPLETED` i think there will be an issue for existing records. I guess night shelter is just started, not sure if any application has come to QA complete stage

gio-karunakaran: I've kept RESOLVED as 3 for now, can changes this back to 2 in the future if it will have no implications with data already in the assessment_store db 👍🏾 I will also update assessment frontend to consider QA_complete as a separate workflow rather than a flag.

db/migrations/versions/6277d910baa8_.py
RamuniN, line 1: Have you tested this migration file? I am getting error while trying to upgrade
```
Exception has occurred: DataError
(psycopg2.errors.InvalidTextRepresentation) invalid input value for enum flagstatus: "QA_COMPLETED"

[SQL: ALTER TABLE flag_update ALTER COLUMN status TYPE flagstatus USING status::text::flagstatus]
(Background on this error at: https://sqlalche.me/e/14/9h9h)
psycopg2.errors.InvalidTextRepresentation: invalid input value for enum flagstatus: "QA_COMPLETED"
```

db/migrations/versions/6277d910baa8_.py
```diff
@@ -0,0 +1,74 @@
...
+    op.execute(
+        "CREATE TYPE flagstatus AS ENUM ('RAISED', 'STOPPED', 'RESOLVED')"
+    )
+    op.execute(
+        "ALTER TABLE flag_update ALTER COLUMN status TYPE flagstatus USING status::text::flagstatus"
+    )
+    op.execute(
+        "ALTER TABLE assessment_flag ALTER COLUMN latest_status TYPE flagstatus USING latest_status::text::flagstatus"
+    )
+
```
RamuniN, line 53: I think you are missing `op.execute("DROP TYPE flagstatus_old")` line at the end of upgrde function

db/migrations/versions/6277d910baa8_.py
```diff
@@ -0,0 +1,74 @@
...
+    op.execute("ALTER TYPE flagstatus RENAME TO flagstatus_old")
+    op.execute(
+        "CREATE TYPE flagstatus AS ENUM ('RAISED', 'STOPPED', 'QA_COMPLETE', 'RESOLVED')"
+    )
+    op.execute(
+        "ALTER TABLE flag_update ALTER COLUMN status TYPE flagstatus USING status::text::flagstatus"
+    )
+    op.execute(
+        "ALTER TABLE assessment_flag ALTER COLUMN latest_status TYPE flagstatus USING latest_status::text::flagstatus"
+    )
```
RamuniN, line 73:
```suggestion
    with op.get_context().autocommit_block():
        op.execute("ALTER TYPE flagstatus ADD VALUE 'QA_COMPLETED'")
```
Are you still working on downgrade function? Looks you haven't updated it yet

db/migrations/versions/6277d910baa8_.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    public_pg_trgm = PGExtension(schema="public", signature="pg_trgm")
```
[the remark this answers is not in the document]
gio-karunakaran: have removed this 👍🏾

openapi/api.yml
```diff
@@ -932,3 +932,75 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: 'components.yml#/components/schemas/Error'
+              example:
+                code: 401
+                status: 'error'
+                message: 'Could not mark application as QA_COMPLETE'
+
+  '/qa_complete/{application_id}':
```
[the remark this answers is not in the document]
gio-karunakaran: this will only get qa info for the old qa flag, once we remove that we'll need to use this new endpoint that searches if there is a qa_complete record for an application 👍🏾

db/migrations/versions/6277d910baa8_.py
gio-karunakaran: it works for me, but I had to drop my old assessment table and rebuild + re-seed completely

db/migrations/versions/6277d910baa8_.py
gio-karunakaran: ahh great spot thank you! have added this now

db/migrations/versions/6277d910baa8_.py
gio-karunakaran: I missed this line in the downgrade function: op.execute("DROP TYPE flagstatus_old") but have added now and should be all done 👍🏾

db/migrations/versions/6277d910baa8_.py
gio-karunakaran: have now separated the migration file to 2 seperate migration files.

RamuniN (dismissed):
👍

PR #223, opened 2023-07-11
fix integer value rendering

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3097
https://dluhcdigital.atlassian.net/browse/FS-3070

### Change description
- fix integer value rendering

### How to test

### Screenshots of UI changes (if applicable)
image


config/mappings/assessment_mapping_schema.py
```diff
@@ -78,6 +78,7 @@
                     # covers:
                     # new add-another component with children support
                     # see https://github.com/communitiesuk/digital-form-builder/pull/161
+                    "integer",
```
tferns, line 81: Maybe we could leave a comment
```suggestion
                    "integer",
                    # covers:
                    # numberField (when we have a number input we want as a whole number)
```

tferns (dismissed):

rayhan15243 (dismissed):

config/mappings/assessment_mapping_schema.py
RamuniN: Done

rayhan15243 (dismissed):

tferns (approved):

PR #246, opened 2023-07-21
BAU: Return empty list if no associated tags

tferns:
- Return an empty list if no tags, so we get a list and a `200` response
- If we return `None`, it implicitly gives us a `204`

https://communities-govuk.slack.com/archives/C02PHBER287/p1689942993480279


Adam-W1 (approved):

gio-karunakaran (dismissed):
🤌🏾

gio-karunakaran (approved):
👌🏾

PR #261, opened 2023-08-07
Added tags relation in db models

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3269

## Description
- Tags are now retrieved from application overviews API (no need to make separate API call to fetch tags) to improve performance
- Added tags relation in db models

### Before (3.94 seconds to load 50 applications)
image
### After (1.73 seconds to load 50 applications)
image


Adam-W1 (approved):
lgtm

PR #269, opened 2023-08-15
FS-3337 - Split out AWS test/deploy from PaaS

robk-dluhc:
https://dluhcdigital.atlassian.net/browse/FS-3337

### Change description
Move AWS test/deploy out from the PaaS one.
Note this is currently pulling a branch of workflows - revert this before merging.

- [ ] Unit tests and other appropriate tests added or updated
- [ ] README and other documentation has been updated / added (if needed)
- [ ] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test


.github/workflows/deploy-aws.yml
aaronwilliamsv1, line 1: How, exactly, does this file run? There's no call or reference to it, it's just isolated now

robk-dluhc: I thought all the things under workflows get consolidated automatically?

.github/workflows/deploy.yml
aaronwilliamsv1, line 1: The way it currently runs, the script packages all-in-one a unit test, then deploy, then tests against the deploy. We need to split that out, as in the original run the test_and_deploy step does cloudfoundry all in one, and we need to run the separated out versions

.github/workflows/deploy-aws.yml
aaronwilliamsv1: Hmmm...now that you mention it, it will trigger automatically, because there ARE no trigger restrictions. Never mind, carry on

PR #277, opened 2023-09-12
FS-3440 log location extraction errors to sentry

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3440

### Description
Currently location extraction errors are not logged in sentry during application import
- Log sentry error when location keys are not valid or invalid postcode provided

### How to test
- Submit an application with invalid postcode in the projection information section
- Wait for the automatic import to trigger (send & receive message (~ 60sec))
- Observe that import should be successful & location errors are logged to sentry (or in assessment container logs if testing locally)


tferns (approved):
👍 LGTM

PR #278, opened 2023-09-12
FS-3434: Add missing question to test json

tferns:
- Add missing question to test json, so we can test properly


ramsharma-prog (approved):
lgtm

PR #279, opened 2023-09-12
FS-3436: Change text of question

tferns:
Simply changing text according to the ticket here: https://dluhcdigital.atlassian.net/browse/FS-3436

image


ramsharma-prog (approved):
lgtm

PR #280, opened 2023-09-12
FS-3433: Change numberField to textField to render as raw number

tferns:
- Change numberField to textField to render as raw number

https://dluhcdigital.atlassian.net/browse/FS-3433

image


ramsharma-prog (approved):
lgtm

PR #281, opened 2023-09-12
FS-3435: Fix type of revenue cost & import

tferns:
- Import revenue costs as number instead of string (mimicking real app)
- Change type to `numberField` as `currency` isn't a valid type, it's derived 

image


RamuniN (approved):
👍

PR #284, opened 2023-09-15
FS-3441 added yelp hook for secrets

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3441

### Change description
- Added pre-commit 'Yelp/detect-secrets' to prevent committing of secrets to code base
- Ignore false positives by marking them with `pragma: allowlist secret`
- disabled the plugin `HexHighEntropyString` which tend to detect lot of false positives

### How to test
- Add a secret/API/password to any of the file type
  For eg. `API_KEY= "125fdfbdjhbj5989"`
- stage the file & commit it.
- Notice the commit is failed with errors.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -2,6 +2,8 @@ FLASK_APP=app.py
 FLASK_ENV=development
 FLASK_RUN_PORT=5000
 FLASK_RUN_HOST=localhost
+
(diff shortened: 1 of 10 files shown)


Adam-W1 (approved):
Looks good, the earlier we catch it the better 👍

PR #291, opened 2023-10-04
FS-3575: Add is_withdrawn flag and update queries

tferns:
### Change description

- Adds a column `is_withdrawn` to assessment records, which acts as a flag to determine whether a record is marked as withdrawn
- The queries have been updated to not pull back any withdrawn applications, so we essentially "hide" it from the user interface
- This is done to match a support query to "withdraw" an application without losing the data, it would be good to have an option in the UI for this though to allow lead assessors to do this

---

- [x] README and other documentation has been updated / added (if needed)
- [x] Commit messages are meaningful and follow good commit message guidelines

### DIFF
--- a/db/migrations/versions/~2023_10_04_1057-342b7a05b923_.py
+++ b/db/migrations/versions/~2023_10_04_1057-342b7a05b923_.py
@@ -1,0 +1,29 @@
+"""empty message
+
+Revision ID: 342b7a05b923
+Revises: 222a1c3b6321
+Create Date: 2023-10-04 10:57:01.282981
+
+"""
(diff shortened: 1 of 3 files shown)


Adam-W1 (approved):
LGTM 👍 
 do we need to add the 'is_withdrawn' to the application models in the other services as part of this?

tferns:
> LGTM 👍 do we need to add the 'is_withdrawn' to the application models in the other services as part of this?

Yeah - I'm doing a change at the moment moving the round/fund into utils so we can share it across services, I'll include this!

PR #296, opened 2023-10-12
Fs 3537 update import script configuration for cyp r1

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3537

### Description
- Updated fund_round_data_key_mappings for CYP
- updated application JSON template for CYP
- Update the seeding script for CYP
- Test the updated import script & make sure all fields are mapped accordingly
- Disable automatic import of application when mapping config is missing for a round(or not configured yet for a new round )

### DIFF
--- a/db/models/assessment_record/assessment_records.py
+++ b/db/models/assessment_record/assessment_records.py
@@ -86,9 +86,15 @@ class AssessmentRecord(BaseModel):
     # These are defined as column_properties not as hybrid_property due to performance
     # Using column_property below forces the json parsing to be done on the DB side which is quicker than in python
     organisation_name = column_property(
-        func.jsonb_path_query_first(
-            jsonb_blob,
-            '$.forms[*].questions[*].fields[*] ? (@.key == "opFJRm").answer',
+        func.coalesce(
+            func.jsonb_path_query_first(
+                jsonb_blob,
+                '$.forms[*].questions[*].fields[*] ? (@.key == "opFJRm").answer',
+            ),
+            func.jsonb_path_query_first(
+                jsonb_blob,
+                '$.forms[*].questions[*].fields[*] ? (@.key == "JbmcJE").answer',
+            ),
         )
     )
     local_authority = column_property(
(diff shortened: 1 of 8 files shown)


db/schemas/schemas.py
```diff
@@ -36,6 +36,8 @@ class Meta:
     organisation_name = String()
     funding_type = String()
     local_authority = String()
+    cohort = String()
+    is_project_regional = Boolean()
```
Adam-W1, line 40: does this need adding to the returned payload for the frontend (like in the api.yml)?

RamuniN: you meant in the components.yml. It is defined as json object there so not required to add https://github.com/communitiesuk/funding-service-design-assessment-store/blob/d69b7677337bb02a666572a98800a7a34dac66f7/openapi/components.yml#L22-L23

db/models/assessment_record/assessment_records.py
```diff
@@ -86,9 +86,15 @@ class AssessmentRecord(BaseModel):
     # These are defined as column_properties not as hybrid_property due to performance
     # Using column_property below forces the json parsing to be done on the DB side which is quicker than in python
     organisation_name = column_property(
-        func.jsonb_path_query_first(
-            jsonb_blob,
-            '$.forms[*].questions[*].fields[*] ? (@.key == "opFJRm").answer',
+        func.coalesce(
```
Adam-W1, line 89: As we have discussed, this approach is less than ideal going forward. I think the next best thing is adding it to the assessment_mapping_fund_round so at least we are maintaining the pointers to json fields in one place. One to add to tech debt i think - what do you think?

db/schemas/schemas.py
Adam-W1: ah okay, I think that's probably outdated as its missing some schema on the retuned assessment (for example i think 'applicationId' id actually 'application_id')

db/models/assessment_record/assessment_records.py
Adam-W1: yeah I understand, i think we just make a note that this is another area that needs refactoring 👍

Adam-W1 (approved):
LGTM 👍

PR #301, opened 2023-10-19
use correct revenue field keys

Adam-W1:
https://dluhcdigital.atlassian.net/browse/FS-3661

### DIFF
--- a/config/mappings/assessment_mapping_fund_round.py
+++ b/config/mappings/assessment_mapping_fund_round.py
@@ -123,7 +123,6 @@ fund_round_data_key_mappings = {
         "funding_two": ["XsAoTv", "JtBjFp"],
     },
     "CYPR1": {
-        "field_type": "multiInputField",
         "location": "rmBPvK",
         "asset_type": None,
         "funding_one": None,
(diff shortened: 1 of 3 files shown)


config/mappings/assessment_mapping_fund_round.py
```diff
@@ -127,7 +127,7 @@
         "location": "rmBPvK",
         "asset_type": None,
         "funding_one": None,
-        "funding_two": ["qwktzL", "JizgZP"],  # only revenue funding for cyp
+        "funding_two": ["JXKUcj", "OnPeeS"],  # only revenue funding for cyp
```
RamuniN, line 130: you may need to remove this `"field_type": "multiInputField",` from CYP as it's not multiInputField anymore

Adam-W1: good catch - perhaps we should specify on that field key, something like 'funding_field_type''

tferns (approved):
lgtm!

scripts/data_updates/FS-3661-total-funding-requested-cyp.py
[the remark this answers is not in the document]
Adam-W1: yeah thats right, just once. you should be able to try it locally and see the values update: 
docker exec -it {container id} python -m scripts.data_updates.FS-3661-total-funding-requested-cyp

RamuniN (approved):

PR #303, opened 2023-10-20
add support for both capitilised key and non capitilized field types

Adam-W1:
cyp config patches

https://dluhcdigital.atlassian.net/browse/FS-3648
https://dluhcdigital.atlassian.net/browse/FS-3646

### DIFF
--- a/api/routes/subcriterias/get_sub_criteria.py
+++ b/api/routes/subcriterias/get_sub_criteria.py
@@ -201,6 +201,7 @@ def format_add_another_component_contents(
 
         table = []
         for component_id, column_config in table_config.items():
+            frontend_format = None
             title = column_config["column_title"]
             answers = component_id_to_answer_list.get(component_id)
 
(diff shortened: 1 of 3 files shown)


api/routes/subcriterias/get_sub_criteria.py
```diff
@@ -171,6 +171,7 @@ def deprecated_sort_add_another_component_contents(
         "RadioField": "text",
         "textField": "text",
         "MonthYearField": "monthYearField",
+        "monthYearField": "monthYearField",
```
RamuniN, line 174: Is this duplicating above above field?

Adam-W1: yes, it was sensitive to the case, so i specified support for both cases. This has now been superseded by matching without case sensitivity in the function, see https://github.com/communitiesuk/funding-service-design-assessment-store/commit/a5ad86ce68d39ca9d4c2c1a07a6cd1b0f5f12b7f

RamuniN (approved):
LGTM

PR #363, opened 2023-11-29
bug-fix/revert-application-id-to-none-for-get-commets-endpoint

ramsharma-prog:
bug-fix/revert-application-id-to-none-for-get-commets-endpoint

### DIFF
--- a/api/routes/score_routes.py
+++ b/api/routes/score_routes.py
@@ -9,7 +9,7 @@ from flask import request
 
 
 def get_score_for_application_sub_criteria(
-    application_id: str,
+    application_id: str = None,
     sub_criteria_id: str = None,
     score_history: bool = False,
 ) -> List[Dict]:
--- a/db/queries/comments/queries.py
+++ b/db/queries/comments/queries.py
@@ -15,7 +15,7 @@ from sqlalchemy import select
 
 
 def get_comments_for_application_sub_crit(
-    application_id: str,
+    application_id: str = None,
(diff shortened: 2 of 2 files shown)


rayhan15243 (dismissed):

rayhan15243 (dismissed):

ramsharma-prog:
> Couple of comments:
> 
> * If this is a bug fix is there a unit test that needs adding to make sure the bug doesn't come back?
> * Can you add the ticket ref etc in the description please as it's not got much info at the moment.
> 
> Code looks fine tho! :)

It wasn't a bug. The application_id was set to None originally but i changed it, though the endpoint /comment will always have the application_id but that is not the case so reverted back. @srh-sloan

AbuNavsa (dismissed):
lgtm

tests/test_db_function.py
```diff
@@ -222,6 +224,12 @@ def test_get_comments(seed_application_records):
     )
     assert len(comment_metadata_no_theme) == 3
 
+    # test without application_id
+    comment_metadata_for_comment_id = get_comments_for_application_sub_crit(
+        application_id=None, comment_id=comment_metadata["id"]
```
srh-sloan, line 229: Probably you need to not pass in application_id at all to replicate the error?

srh-sloan (commented):
One tweak

srh-sloan (approved):
👍

PR #395, opened 2024-01-15
Bau increase test cpu

robk-dluhc:
Increase the CPU of assessment-store so it doesn't max out when tests are run.

### Change description

- [n/a] Unit tests and other appropriate tests added or updated
- [n/a] README and other documentation has been updated / added (if needed)
- [X] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test


copilot/fsd-assessment-store/manifest.yml
```diff
@@ -57,6 +57,8 @@ environments:
   test:
     deployment:
       rolling: 'recreate'
+    cpu: 1024
+    memory: 2048
```
gidsg, line 61: Does it make sense to be running with more resources in test than in production? Should we change the default for the app instead?

robk-dluhc: The thinking was that test runs e2e testing frequently.  UAT and Prod are the same.  I'm happy to change them all to be higher if that seems better.

aaronwilliamsv1: Dev: 
image


Test:
image


UAT:
image


Prod:
image

aaronwilliamsv1: Test is regularly pushing the limits of it's resources, I suspect due to the E2E tests that run on assessment

gidsg: Seems a bit weird, e.g. if we are running performance tests on pre-prod environments it will behave differently on production, so I'd rather level up the default for that reason.

aaronwilliamsv1: This is on the test environment, which will, by it's very nature, be stressed MORE than production environments. Because you're running tests against it, it will have more connections, use more resources, stress the CPU more often, etc. Development may end up this way as well, depending on how we end up architecting the dev environment, but the stress of this is usually taken away by running all of that functionality locally. That's why it goes test -> uat -> prod

aaronwilliamsv1: If we level up the default, in my opinion we're wasting resources on environments that don't need buffing, but I'm not fussed either way - we can either up the default, or test specifically based on the load patterns

aaronwilliamsv1 (approved):
Let's get it in, revert it if it's deemed unnecessary

copilot/fsd-assessment-store/manifest.yml
robk-dluhc: For now, we'll get this in.  Resolving conversation.  If required we can extend this increase to the other envs.

PR #401, opened 2024-01-23
FS-3279 filter null values in address field

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3279
https://dluhcdigital.atlassian.net/browse/FS-3721

### Description
- Filter null values in the address field
- Extract plain text from `freeText` type and display in the exported application .

### Screenshots
image
image

### DIFF
--- a/db/queries/assessment_records/queries.py
+++ b/db/queries/assessment_records/queries.py
@@ -991,7 +991,14 @@ def get_export_data(
                             title = form_fields[field["key"]][language][
                                 "title"
                             ]
-                            answer = field["answer"]
+
+                            # filter 'null' values from the address field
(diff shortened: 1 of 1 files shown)


Adam-W1 (approved):

db/queries/assessment_records/queries.py
```diff
@@ -991,8 +992,23 @@ def get_export_data(
                             title = form_fields[field["key"]][language][
                                 "title"
                             ]
-                            answer = field["answer"]
+
```
Adam-W1, line 995: does this have test coverage?

RamuniN: I think no

PR #435, opened 2024-03-04
FS-4200-remove-html-tags-from-export-applicat-information

ramsharma-prog:
### Ticket
https://dluhcdigital.atlassian.net/jira/software/c/projects/FS/boards/50?selectedIssue=FS-4200

### Description
Add the correct question title to cof-eoi


rayhan15243:
@ramsharma-prog How about the other cosmetic issues with that export? The ticket was more than just removing the html tags.

rayhan15243 (approved):

PR #498, opened 2024-09-20
update manifest

srh-sloan:
update manifest


sfount (approved):
Is it worth associating all of these with a Jira ticket just in case this sharp edge + the context and motivation for these changes needs explaining to future con figurers?

PR #499, opened 2024-09-23
change logging level

Adam-W1:
Use a WARNING rather than an ERROR when a fund-round uses the default scoring system.

Additionally..
Warn is deprecated so fix an instance of this:
image


srh-sloan (approved):
LGTM 🚀

hamzabinkhalid (approved):
LGTM

PR #500, opened 2024-09-23
FS-4446: Consistent CSV File Columns

suegarner:
[FS-446](https://dluhcdigital.atlassian.net/browse/FS-4446)

### Change description
Rearranged the default order of the columns so that Score comes directly after Date submitted regardless whether there is a score entered or not. Previously Score and Score Subcriteria could be swapped depending on if the first row had been scored yet)

- [ ] ~~README and other documentation has been updated / added (if needed)~~
- [x] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
Open OUTPUT_TRACKER report for a window that has applications but no scores. Open OUTPUT_TRACKER for a window that has applications and are all scored. The columns should be in the same order for both reports.

### Screenshots of UI changes (if applicable)

Before fix (COF OUTPUT_TRACKER report):
image
image

NSTF OUTPUT_TRACKER report:
image
image

[FS-446]: https://dluhcdigital.atlassian.net/browse/FS-446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/queries/assessment_records/queries.py
+++ b/db/queries/assessment_records/queries.py
@@ -576,8 +576,8 @@ def get_assessment_records_score_data_by_round_id(round_id, selected_fields=None
     default_fields = [
         "Application ID",
         "Short ID",
-        "Score Subcriteria",
         "Score",
+        "Score Subcriteria",
         "Score Justification",
         "Score Date",
         "Score Time",
--- a/tests/test_db_function.py
+++ b/tests/test_db_function.py
@@ -553,10 +553,10 @@ def test_output_tracker_columns_remain_same_for_scored_and_unscored_reports(seed
             }
         },
     )
-  
-    no_scores_cols = list(no_scores['en_list'][0].keys())
-    with_scores_cols = list(with_scores['en_list'][0].keys())[:len(no_scores_cols)]
-    assert  no_scores_cols == with_scores_cols
+
+    no_scores_cols = list(no_scores["en_list"][0].keys())
+    with_scores_cols = list(with_scores["en_list"][0].keys())[: len(no_scores_cols)]
+    assert no_scores_cols == with_scores_cols
 
 
 @pytest.mark.apps_to_insert([test_input_data[4]])  # taken from assessment store for cof r4w1
--- a/tests/test_db_function.py
+++ b/tests/test_db_function.py
@@ -507,6 +507,58 @@ def test_output_tracker_with_no_scores_data(seed_application_records, mocker):
     assert data["en_list"][0]["Risks to your project (document upload)"] == "sample1.doc"
 
 
+@pytest.mark.apps_to_insert([test_input_data[0]])
+def test_output_tracker_columns_remain_same_for_scored_and_unscored_reports(seed_application_records, mocker):
+    mocker.patch(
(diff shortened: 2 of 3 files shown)


db/queries/assessment_records/queries.py
```diff
@@ -576,8 +576,8 @@ def get_assessment_records_score_data_by_round_id(round_id, selected_fields=None
     default_fields = [
```
sfount, line 576: It looks like there's one fund that specifies its own `score_fields` which if I understand correctly will override these "default" fields and their order. As its only one it might be worth making sure the score order is consistent there too https://github.com/communitiesuk/funding-service-design-assessment-store/blob/511558a60544efcee9074611141bba423e391b08/config/mappings/assessment_mapping_fund_round.py#L261-L268

db/queries/assessment_records/queries.py
```diff
@@ -576,8 +576,8 @@ def get_assessment_records_score_data_by_round_id(round_id, selected_fields=None
...
         "Application ID",
         "Short ID",
-        "Score Subcriteria",
         "Score",
+        "Score Subcriteria",
```
sfount, line 580: I wonder if we can extend the `test_output_tracker_data()` to write a failing test that can then guard against regression on the order of the scoring columns based on if its scored or not

I'm happy to pair on this if helpful but maybe it could be something along the lines of:
- call the export data method with no scores in the database 
- add scores to the database
- call the export data method
- assert that the `.keys()` of the dictionaries in both returned lists start with the same thing

Strangely it looks like these tests actually use the database rather than fixtures so there may be some sharp edges but hopefully theres something to go on.

I don't think that this needs to block this PR, especially if appeasing the tests takes longer than reasoning about the change!

sfount (commented):
Looks good, a subtle difference but looks like one that could crop up for users at anytime!

The description on the PR is really helpful context on why that order might have been shifted, if that could be added to the commit description itself that should help any future devs outside of GitHub looking at why that might have changed.

tests/test_db_function.py
```diff
@@ -507,6 +507,58 @@ def test_output_tracker_with_no_scores_data(seed_application_records, mocker):
...
+        {
+            "OUTPUT_TRACKER": {
+                "form_fields": {
+                    "KAgrBz": {"en": {"title": "Project name"}},
+                }
+            }
+        },
+    )
+
+    with_scores_cols = list(with_scores["en_list"][0].keys())[:-5]
```
sfount, line 557: Could the number of columns picked at the end use the length of the `no_scores_cols` keys? Assuming with score will always be longer than without scores that would just mean this wouldn't break if more or less (consistent and shared!) columns are added at the beginning

sfount (commented):
PR looks really good and well covered! One small suggestion for making the coverage more dynamic but I think this is good to undraft and review/ approve. Great fix!

nuwan-samarasinghe (approved):
LGTM

PR #502, opened 2024-09-24
Increase number of process workers for assessment-store

ksp37-dluhc:
Assessment-store can sometimes receive multiple API requests for the same high-level request originating from assessment. It would be beneficial to have multiple workers to be able to process concurrent requests. 

### Change description
_A brief description of the pull request_

- [ ] Unit tests and other appropriate tests added or updated
- [ ] README and other documentation has been updated / added (if needed)
- [ ] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
-web: gunicorn --worker-class uvicorn.workers.UvicornWorker -c run/gunicorn/devtest.py wsgi:app
+
+web: gunicorn --worker-class uvicorn.workers.UvicornWorker -c run/gunicorn/devtest.py --workers 3 wsgi:app


Procfile
```diff
@@ -1 +1,2 @@
-web: gunicorn --worker-class uvicorn.workers.UvicornWorker -c run/gunicorn/devtest.py wsgi:app
+
+web: gunicorn --worker-class uvicorn.workers.UvicornWorker -c run/gunicorn/devtest.py --workers 3 wsgi:app
```
nuwan-samarasinghe, line 2: @ksp37-dluhc  why do we need to increase the amount of workers ?

gidsg (approved):
LGTM 👍

gidsg:
I think this will be an improvement, but it should not be actually one request per worker i.e. as described in https://stackoverflow.com/a/48543832 and https://github.com/communitiesuk/funding-service-design-utils/blob/024fa53fe1be7f5d6a6ccf0f35f90febc1c9c69c/fsd_utils/gunicorn/config/devtest.py#L76C1-L76C19

ksp37-dluhc:
> I think this will be an improvement, but it should not be actually one request per worker i.e. as described in https://stackoverflow.com/a/48543832 and https://github.com/communitiesuk/funding-service-design-utils/blob/024fa53fe1be7f5d6a6ccf0f35f90febc1c9c69c/fsd_utils/gunicorn/config/devtest.py#L76C1-L76C19

Isn't that only applicable if we're using gevent / greenlet workers?

nuwan-samarasinghe (approved):
LGTM

Procfile
ksp37-dluhc: Some of the routes in Assessment involve multiple (sometimes expensive) requests to the Assessment Store. So with a single worker, Assessment Store may only be able to process the requests sequentially. But with multiple workers it can process multiple requests at the same time (and so reduce the page load time in Assessment).

PR #534, opened 2024-11-25
Turn off recreate deploys in test environment

samuelhwilliams:
### Change description
This is because the `recreate` deployment method leads to downtime, which can cause tests to flake. It is not any faster than a normal deployment, and causes tests to break. I am not sure why it was chosen to begin with.

https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#deployment-rolling


MarcUsher (approved):
👍

